Issue #10, opened 2018-03-27
ddmrp_production_equivalent


max3903:
This module will allow you to use equivalent products when creating a manufacturing order.

Odoo will assume that products in the same product category or subcategories are equivalent.

# Configuration

* Go to Manufacturing > Master Data > Products
* Create the following products:

Product | Priority | Product Category | Buffer Level (Netflow)
-- | -- | -- | --
Server | 10 | Servers | red
Part A | 10 | Parts | red
Part B | 20 | Parts/Sub 1 | red
Part C | 30 | Parts/Sub1/Sub2 | green
Part D | 40 | Parts/Sub1/Sub2/Sub3 | green
Part E | 50 | Parts | green
Part F | 20 | Office Supplies | green

* Go to Manufacturing > Master Data > Bill of Materials
* Create a BOM for 1 Server with:

Component | Quantity | Use Equivalences | Non-Equivalent Products
-- | -- | -- | --
Part A | 1 | True | Part C

# Usage

* Go to Manufacturing
* Create a manufacturing order for 15 Servers
* The manufacturing order will be consuming 15 Part D because:
  * Part A buffer is red
  * Part B is equivalent but the buffer is red
  * Part C is excluded
  * Part D has a higher priority than E
  * Part F is in a different product category

LoisRForgeFlow:
related PR is merged.

PR #3, opened 2018-03-09
[WIP] Add ddmrp from Eficent repository

guewen:
* Extract the addon from https://github.com/Eficent/ddmrp
* Migrate ddmrp to version 11.0
* Note: working on https://github.com/Eficent/ddmrp/pull/47 until ready, then I'll update this one for inclusion in OCA

The other addons of the repository will need to be extracted the same way.

Require #2 for the build

PR #11, opened 2018-03-30
[11.0][ADD] ddmrp_production_equivalent module

agyamuta:
This module will allow you to use equivalent products when creating a manufacturing order.

#10 

Depends on:
- [ ] #3


ddmrp_production_equivalent/models/mrp_production.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def _get_product_equivalent(self, bom_line, requested_qty):
+        # get all the other products in the same product category
+        p_obj = self.env['product.product']
+        product_ids = p_obj.search([
+            ('categ_id', 'child_of', bom_line.product_id.categ_id.id)],
+            order='priority asc')
+        # exclude the non-equivalent parts listed in the BOM line
+        product_ids = product_ids - p_obj.browse(
+            [p.id for p in bom_line.nonequivalent_product_ids])
```
guewen, line 18: You shouldn't need to browse nor to iterate to get the ids:

```python
products -= bom_line.nonequivalent_product_ids
```

ddmrp_production_equivalent/models/mrp_production.py
```diff
@@ -0,0 +1,45 @@
...
+from odoo import models
+
+
+class MrpProduction(models.Model):
+    _inherit = 'mrp.production'
+
+    def _get_product_equivalent(self, bom_line, requested_qty):
+        # get all the other products in the same product category
+        p_obj = self.env['product.product']
+        product_ids = p_obj.search([
```
guewen, line 13: please rename `product_ids` to `products`, the variable doesn't contain ids but records

ddmrp_production_equivalent/models/mrp_production.py
```diff
@@ -0,0 +1,45 @@
...
+            ('categ_id', 'child_of', bom_line.product_id.categ_id.id)],
+            order='priority asc')
```
guewen, line 15: Please extract this search in another method so it can be overidden.

`priority asc, id asc` so if you have 2 products with the same priority, you'll always have the same one and not a flip-flop which can be surprising

ddmrp_production_equivalent/models/mrp_production.py
```diff
@@ -0,0 +1,45 @@
...
+                replace_product = False
+            else:
+                replace_product = True
+        elif not bom_line.use_equivalences:
+            replace_product = False
+        else:
+            replace_product = True
+            product_equivalent = self._get_product_equivalent(bom_line,
+                                                              line_data['qty'])
+        if replace_product and product_equivalent:
```
guewen, line 41: `product_equivalent` is defined only when the `else` condition path is taken, the rest of the time, it'll crash.

Seems to me that the following algorithm would make sense according to my understanding:

```python
if not bom_line.use_equivalences:
    return sm
if (bom_line.orderpoint_id.planning_priority_level == '3_green' and
        bom_line.product_id.qty_available > line_data['qty']):
    return sm
product_equivalent = self._get_product_equivalent(bom_line, line_data['qty'])
if product_equivalent:
    sm.write({
        'price_unit': product_equivalent.standard_price,
        'product_id': product_equivalent.id,
    })
return sm
```

ddmrp_production_equivalent/models/mrp_production.py
```diff
@@ -0,0 +1,45 @@
...
+            # update stock move product id and  price unit
+            sm.product_id = product_equivalent.id
+            sm.price_unit = product_equivalent.standard_price
```
guewen, line 44: This is issuing two different `UPDATE` in database which is inefficient. Replace by a call to write

guewen (changes requested):

max3903:
Superseeded by #12

PR #12, opened 2018-04-17
[ADD] ddmrp_production_equivalent module

max3903:
#10 
Depends on #3
Superseeds #11


guewen (approved):

ddmrp_production_equivalent/models/mrp_production.py
```diff
@@ -0,0 +1,42 @@
...
+        return [('categ_id', 'child_of', bom_line.product_id.categ_id.id)]
+
+    def _get_product_equivalent(self, bom_line, requested_qty):
+        # get all the other products in the same product category
+        p_obj = self.env['product.product']
+        products = p_obj.search(self._get_product_search_criteria(bom_line),
+                                order='priority asc, id asc')
+        # exclude the non-equivalent parts listed in the BOM line
+        products -= bom_line.nonequivalent_product_ids
+        for product in products:
```
moylop260: <s>Could you help me to understand this loop?

The `break` sentence finish the loop but I didn't see any assignation or computation after or before... then what is the reason of run this loop without any action?</s>
Forget it, you are using the `product` variable (last value of the loop) to return it.

moylop260: How you are avoiding a `NameError: name 'product' is not defined` error?

I mean,
```python
products = []
for product in products:
    pass
print(product)
```
```bash
$ python3.5 test.py
Traceback (most recent call last):
  File "test.py", line 4, in <module>
    print(product)
NameError: name 'product' is not defined
```

max3903:
@moylop260 Good catch! Thanks. With our data set, products was never empty.

ddmrp_production_equivalent/models/mrp_production.py
max3903: @moylop260 Good catch! Thanks. With our data set, products was never empty.

LoisRForgeFlow:
why did you merge this without the dependencies merged?

PR #13, opened 2018-04-20
[IMP] Log message when a component is replaced

max3903:


bodedra (approved):
Code review :+1:

ddmrp_production_equivalent/models/mrp_production.py
```diff
@@ -36,8 +36,12 @@ def _generate_raw_move(self, bom_line, line_data):
         product_equivalent = self._get_product_equivalent(bom_line,
                                                           line_data['qty'])
         if product_equivalent:
+            body = _('[%s] %s has been replaced by [%s] %s.' %
+                     sm.product_id.default_code, sm.product_id.name,
```
moylop260: What about use `name_get()[0][1]` instead of a manual `[%s] %s` in order to get the full name if there is an inherited method changing this one (as product variants case or product_brand module)?

max3903:
@moylop260 Fixed! Thanks.

moylop260 (approved):
Thanks 😍

ddmrp_production_equivalent/models/mrp_production.py
```diff
@@ -36,8 +36,12 @@ def _generate_raw_move(self, bom_line, line_data):
         product_equivalent = self._get_product_equivalent(bom_line,
                                                           line_data['qty'])
         if product_equivalent:
+            body = _('%s has been replaced by %s.' %
+                     sm.product_id.name_get()[0][1],
+                     product_equivalent.name_get()[0][1])
```
agyamuta: Tried to run this code and got this error.
*** TypeError: not enough arguments for format string

Would suggest the following lines

            body = _('%s has been replaced by %s.' %
                     (sm.product_id.name_get()[0][1],
                     product_equivalent.name_get()[0][1]))

agyamuta (changes requested):

agyamuta (approved):
LGTM!

JordiBForgeFlow:
@max3903 In the future please respect that this repo is actively managed by @lreficent and myself. We establish the policy that only the maintainers of the repo can do merges, as @sbidoul does for OCA/mis-builder. This will help us to better ensure the quality of the repo.

Thank you!

max3903:
@jbeficent Let's be constructive here please. If you want to show leadership, tell me what's wrong here. Do not bring me the "I am the boss" argument, you will just exacerbate frustration.

And by the way, where is your policy documented?

JordiBForgeFlow:
@max3903 we really appreciate your help and I think that we're working well together, as we are doing with stock-rma. But the OCA policies require that at least one of the maintainers approve and merge the PR or a core maintainer does.

This module is also missing tests, which is a failing criteria for merging. And the module was merged with dependencies failing, as pointed out by lois in the PR where it was merged.

In the next days we will be setting up the repo wiki with the governance policies that we would like to set for this project.

max3903:
:+1:

PR #14, opened 2018-04-23
[FIX]In current flow regardless if the product passed or not the last…

agyamuta:
… product from the list is return as the product equivalent.

Correct return product value to False instead of the last product value from the list in case no products passed the given condition.


max3903 (approved):

LoisRForgeFlow:
@agyamuta Can you please check lint? it is making other PRs to fail as well.

max3903:
@lreficent Fixed!

LoisRForgeFlow (approved):

LoisRForgeFlow:
@max3903  Thanks!

PR #16, opened 2018-07-16
[11.0][ADD] ddmrp

LoisRForgeFlow:
This PR includes the main ddmrp module in its latest state superseding https://github.com/OCA/ddmrp/pull/3. The major reasons to move to this PR are:

* The decoupled explosion now uses [mrp_multi_level](https://github.com/OCA/manufacture/tree/11.0/mrp_multi_level) instead of `mrp_mto_with_stock`. This is what changes the data model and adds some extra dependencies. I know this technically can be a pain and we apologize, however:
  * `mrp_mto_with_stock` was merged initially with a lot of bugs and nobody realised, thus I think nobody was using decoupled explosion at that point. I fixed the module and it's now working at OCA/manufacture but it won't be used anymore here on the DDMRP.
  * `mrp_multi_level`: is the right approach, is a very basic traidicional MRP engine, but this way the decoupled explosion happens as it should.
* We have included the option to consider warehouse/plant calendar on the date calculations.
* In this PR we have cleaned the data as discussed with some of you by mail.

@jgrandguillaume @max3903 @guewen @grindtildeath


LoisRForgeFlow:
Small things to complete:
* [x] : update readme
* [x] : rework structure report to make use of the new hook provided by odoo.

grindtildeath (approved):
Reviewed code from 15a15f95e98dfc5a6e18d9905b8ee5e38671ec98 only and looks good.

max3903 (approved):

JordiBForgeFlow (approved):

PR #20, opened 2018-08-02
[11.0][FIX] ddmrp: small fixes on readme and adding new icon.

LoisRForgeFlow:
* wrong module name on roadmap section.
* Alpeph Objects moved from author sections to initial funders.
* Add new, more odooish icon.


JordiBForgeFlow (approved):

pedrobaeza:
Next time please do not delete again the README and put the generic placeholder, as you are generating a very great diff with this. You can leave the old README, and even then, the bot will add the differences with the new auto-generated README from fragments, but the diff will be smaller certainly.

LoisRForgeFlow:
@pedrobaeza Yes,I know but I did that in order to pass travis, as I was having the issue I reported before with de duplicated maintainers target.

pedrobaeza:
Then just remove that line or apply the same patch manually or passing the script locally instead of replacing full file...

LoisRForgeFlow:
@pedrobaeza Oh, good point, that would have been better. Noted for next time!

PR #30, opened 2018-10-30
[11.0][ADD] ddmrp_product_replace

LoisRForgeFlow:
Moved from Eficent/ddmrp.


grindtildeath (approved):
:+1:

JordiBForgeFlow (changes requested):
Small comment. I think that it would be good that the user would be able to view and edit the demand product ids in the buffer. https://github.com/OCA/ddmrp/pull/30/files#diff-909891cd7459d8bacbc530941ace551fR12

LoisRForgeFlow:
@jbeficent Can you check my changes?

carlosdauden (approved):

PR #35, opened 2019-01-21
[11.0][FIX] ddmrp: call _quantity_in_progress with sudo()

HviorForgeFlow:
call _quantity_in_progress with sudo() to avoid ACL errors as is an internal ddmrp engine function.

Error use case:
On instances that use call for tender and/or blanket order management, manufacture users could not use procurement request through stock_orderpoint_manual_procurement due to this.


JordiBForgeFlow (approved):

JordiBForgeFlow:
Fast tracking.

PR #36, opened 2019-01-29
[11.0] ddmrp: performance improvements.

LoisRForgeFlow:
* When the number of MOs associated to a buffer start to grow the calculation of the priority start to be a bottleneck. With this fix it is solved.
* indexing `exclude_from_adu` the ADU calculation gets faster.


JordiBForgeFlow (approved):

HviorForgeFlow (approved):
LGTM :+1: nice fix :smile:

PR #39, opened 2019-02-07
[11.0] ddmrp_history: improve charts.

LoisRForgeFlow:
It can be misleading sometimes, and the execution chart is only meant to focus in the on-hand position.

cc @jbeficent @hveficent

### DIFF
--- a/ddmrp_history/models/stock_warehouse_orderpoint.py
+++ b/ddmrp_history/models/stock_warehouse_orderpoint.py
@@ -1,4 +1,4 @@
-# Copyright 2017 Eficent Business and IT Consulting Services S.L.
+# Copyright 2017-19 Eficent Business and IT Consulting Services S.L.
 #   (http://www.eficent.com)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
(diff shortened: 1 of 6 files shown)


ddmrp_history/models/stock_warehouse_orderpoint.py
```diff
@@ -218,21 +230,23 @@ def stacked(df, categories):
             p = figure(
                 x_range=(dates[0], dates[-1]), y_range=(start_stack, top_y),
                 x_axis_type='datetime')
+            p.sizing_mode = 'stretch_both'
+            p.toolbar.logo = None
 
             p.grid.minor_grid_line_color = '#eeeeee'
             p.patches([x2] * len(areas), [areas[cat] for cat in categories],
                       color=EXECUTION_COLORS, alpha=0.8, line_color=None)
+            date_format = self._get_lang_date_format()
```
carlosdauden: You can remove _get_lang_date_format and use this:
```suggestion
            date_format = self.env['res.lang']._lang_get(self.env.lang).date_format
```
Or self.env.user.lang

carlosdauden (approved):

ddmrp_history/models/stock_warehouse_orderpoint.py
LoisRForgeFlow: Thanks!

AdriaGForgeFlow (approved):
Code review and functional test. LGTM :+1:

PR #40, opened 2019-02-21
[11.0][IMP] ddmrp: improve charts.

LoisRForgeFlow:
Add supply and demand charts.

### DIFF
--- a/ddmrp/models/stock_warehouse_orderpoint.py
+++ b/ddmrp/models/stock_warehouse_orderpoint.py
@@ -4,6 +4,7 @@
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 import logging
+from math import pi
 
 from odoo import api, fields, models, _
 from datetime import datetime, timedelta
@@ -16,6 +17,7 @@ try:
     from bokeh.plotting import figure
     from bokeh.embed import components
     from bokeh.models import Legend, ColumnDataSource, LabelSet
+    from bokeh.models import HoverTool, DatetimeTickFormatter
 except (ImportError, IOError) as err:
     _logger.debug(err)
 
@@ -187,6 +189,7 @@ class StockWarehouseOrderpoint(models.Model):
             p = figure(plot_width=300, plot_height=400,
                        y_axis_label='Quantity')
             p.xaxis.visible = False
+            p.toolbar.logo = None
             red = p.vbar(x=1, bottom=0, top=rec.top_of_red, width=1,
                          color='red', legend=False)
             yellow = p.vbar(x=1, bottom=rec.top_of_red, top=rec.top_of_yellow,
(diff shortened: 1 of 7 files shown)


ddmrp/models/stock_warehouse_orderpoint.py
```diff
@@ -226,6 +229,90 @@ def _compute_ddmrp_chart(self):
...
+                x_demand = list(demand_data.keys())
+                y_demand = list(demand_data.values())
+
+                p = figure(plot_width=500, plot_height=400,
+                           y_axis_label='Quantity', x_axis_type='datetime')
+                p.toolbar.logo = None
+                p.sizing_mode = 'stretch_both'
+                # TODO: # p.xaxis.label_text_font = 'helvetica'
+                p.xaxis.formatter = DatetimeTickFormatter(
+                    hours=["%d %B %Y"], days=["%d %B %Y"], months=["%d %B %Y"],
```
carlosdauden: You can do same that here:
https://github.com/OCA/ddmrp/pull/39

ddmrp/models/stock_warehouse_orderpoint.py
```diff
@@ -226,6 +229,90 @@ def _compute_ddmrp_chart(self):
...
+                       color="firebrick")
+                p.line(
+                    [datetime.today() - timedelta(days=1),
+                     datetime.today() + timedelta(days=rec.order_spike_horizon)],
+                    [rec.order_spike_threshold, rec.order_spike_threshold],
+                    line_width=2, line_dash='dashed')
+
+                unit = rec.product_uom.name
+                hover = HoverTool(
+                    tooltips=[("qty", "$y %s" % unit)],
```
carlosdauden, line 277: You can use .format:
```suggestion
                    tooltips=[("qty", "$y {}".format(unit))],
```

ddmrp/models/stock_warehouse_orderpoint.py
LoisRForgeFlow: Will do now. Thanks

ddmrp/models/stock_warehouse_orderpoint.py
LoisRForgeFlow: See [contributing guidelines](https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#33idioms):

> Prefer % over .format(), prefer %(varname) instead of positional. This is better for translation and clarity.

carlosdauden: Ok, but maybe we should change that in the guidelines:
https://docs.python.org/release/3.1.5/library/stdtypes.html#old-string-formatting-operations

@moylop260 What do you think?

carlosdauden: Sorry, We tested and using format continues to cause problems with translations.
Forget my comment.

AdriaGForgeFlow (approved):
LGTM :+1: once travis is :green_apple:

LoisRForgeFlow:
All comments attended, travis fixed and added tests to increase coverage. Merging.

PR #50, opened 2020-06-29
[13.0][IMP] ddmrp: add dedicated settings and `ddmrp_auto_update_nfp` option.

LoisRForgeFlow:
New option `ddmrp_auto_update_nfp` allows to update NFP after stock moves' state changes that can potentially affect a buffer. This can be significant overhead in systems with a high volume of transactions but it also provides more accurate buffer status without needing to increase the cron job frequency.

@ForgeFlow


jbaudoux (commented):
Some remarks

ddmrp/models/stock_move.py
```diff
@@ -21,3 +21,48 @@ def _merge_moves_fields(self):
         res = super(StockMove, self)._merge_moves_fields()
         res["buffer_ids"] = [(4, m.id) for m in self.mapped("buffer_ids")]
         return res
+
+    def write(self, vals):
+        res = super(StockMove, self).write(vals)
+        if "state" in vals and self.env.company.ddmrp_auto_update_nfp:
```
jbaudoux, line 27: Not all states update should trigger a recomputation

ddmrp/models/stock_move.py
```diff
@@ -21,3 +21,48 @@ def _merge_moves_fields(self):
...
+        out_buffers = self.env["stock.buffer"]
+        for move in self:
+            candidates = buffer_model.search(
+                [
+                    ("product_id", "=", move.product_id.id),
+                    ("location_id", "parent_of", move.location_id.id),
+                ]
+            )
+            for cand in candidates:
+                locations = self.env["stock.location"].search(
```
jbaudoux: better compare parent_path

ddmrp/models/stock_move.py
```diff
@@ -21,3 +21,48 @@ def _merge_moves_fields(self):
...
+        in_buffers = self.env["stock.buffer"]
+        for move in self:
+            candidates = buffer_model.search(
+                [
+                    ("product_id", "=", move.product_id.id),
+                    ("location_id", "parent_of", move.location_dest_id.id),
+                ]
+            )
+            for cand in candidates:
+                locations = self.env["stock.location"].search(
```
jbaudoux: better compare parent_path

jbaudoux (commented):
some remarks

ddmrp/models/stock_move.py
LoisRForgeFlow: which ones don't? I think any status change can potentially affect NFP (note that reservations reduce on-hand from buffer PoV - as qty available not reserved is what is considered on-hand)

ddmrp/models/stock_move.py
```diff
@@ -21,3 +21,48 @@ def _merge_moves_fields(self):
...
+                    ("product_id", "=", move.product_id.id),
+                    ("location_id", "parent_of", move.location_id.id),
+                ]
+            )
+            for cand in candidates:
+                locations = self.env["stock.location"].search(
+                    [("id", "child_of", [cand.location_id.id])]
+                )
+                if move.location_dest_id not in locations:
+                    out_buffers += cand
```
guewen: The whole computation of in_buffers and out_buffers is going to be slow. It generates N moves = N moves * M buffers sql queries.

It might be better to use less `search()` and use relational fields to leverage the prefetch + compare the parent_path for the location tree.

I used this helper on StockLocation in several modules:

```python
    def is_sublocation_of(self, others):
        """Return True if self is a sublocation of at least one other

        It is equivalent to the "child_of" operator, so it includes itself.
        """
        self.ensure_one()
        # Efficient way to verify that the current location is
        # below one of the other location without using SQL.
        return any(self.parent_path.startswith(other.parent_path) for other in others)

```

Example:

```python

for move in self:
    # with a relation, it will do a single sql query to get all the candidates buffers first, instead of one query per move, if we keep most of the buffers as candidates, it should be a clear win
    # buffer_ids would be to add on product.product
    candidates = move.mapped("product_id.buffer_ids").filtered(lambda buffer: move.location_id.is_sublocation_of(buffer.location_id))
    for cand in candidates:
        if not move.location_dest_id.is_sublocation_of(cand.location_id):
            out_buffers += cand
```


Also, you should use sets to store the ids before passing everything to a recordset, because continuously updating a recordet is slow (https://gist.github.com/guewen/a4a2eba5d128b4ed3129866f80571fbd)

ddmrp/models/stock_move.py
```diff
@@ -21,3 +21,48 @@ def _merge_moves_fields(self):
...
+                    ("product_id", "=", move.product_id.id),
+                    ("location_id", "parent_of", move.location_dest_id.id),
+                ]
+            )
+            for cand in candidates:
+                locations = self.env["stock.location"].search(
+                    [("id", "child_of", [cand.location_id.id])]
+                )
+                if move.location_id not in locations:
+                    in_buffers += cand
```
jbaudoux: |= ?

ddmrp/models/stock_move.py
LoisRForgeFlow: Thanks for the insights! I will test and update now.

Side note, have you thought about proposing the `is_sublocation_of` helper to Odoo?

ddmrp/models/stock_move.py
LoisRForgeFlow: Not needed after following @guewen 's approach

ddmrp/models/stock_move.py
LoisRForgeFlow: Not needed after following @guewen 's approach

ddmrp/models/stock_move.py
LoisRForgeFlow: Not needed after following @guewen 's approach

LoisRForgeFlow: Thanks for reviewing! :smiley:

ddmrp/models/stock_move.py
LoisRForgeFlow: Changes done! could you please have a look?

guewen: Nice result with the additional improvement adding the second condition in `filtered`, good catch :)



> Side note, have you thought about proposing the `is_sublocation_of` helper to Odoo?

I didn't, I might propose a more generic `is_child_of()` on master (no way they would merge in 13.0), since it makes sense on any hierarchical record.

guewen (approved):

ddmrp/models/stock_move.py
guewen: @LoisRForgeFlow https://github.com/odoo/odoo/pull/53866

jgrandguillaume (approved):
Ok, that looks really cool now thx !

JordiBForgeFlow (approved):
:+1: tested functionally and technical review. Good!

LoisRForgeFlow:
I have just fixed up "fixup" commits. No code change at all, so restarting merge op:

/ocabot merge minor

ddmrp/models/stock_move.py
LoisRForgeFlow: Great! love it :D

ddmrp/models/stock_move.py
jbaudoux: @LoisRForgeFlow done is not affecting the nfp

PR #52, opened 2020-07-02
[13.0] Add stock_buffer_route

guewen:
It serves the same purpose as stock_orderpoint_route in
OCA/stock-logistics-warehouse, but for DDMRP buffers.

When can select a route that will be used for moves created by a buffer.


stock_buffer_route/tests/test_stock_buffer_route.py
```diff
@@ -0,0 +1,194 @@
...
+                name=name,
+                default_code=default_code,
+                uom_id=cls.env.ref("uom.product_uom_unit").id,
+                company_id=company_id,
+                type="product",
+                **vals
+            )
+        )
+
+    # def test_buffer_route_01(self):
```
guewen: I have to adapt tests for ddmrp

LoisRForgeFlow (approved):
LGTM :+1:

PR #58, opened 2020-07-24
[13.0][IMP] ddmrp_packaging: check MOQ when changing qty multiple

LoisRForgeFlow:
@ForgeFlow


jgrandguillaume (approved):
LGTM even from holidays ;)

jbaudoux (approved):
LGTM

PR #59, opened 2020-07-27
refresh buffer during cron action

JordiBForgeFlow:


AdriaGForgeFlow (approved):

AaronHForgeFlow (approved):
:+1:  Makes sense

ddmrp/models/stock_buffer.py
```diff
@@ -1218,6 +1218,7 @@ def cron_actions(self, only_nfp=False):
         """This method is meant to be inherited by other modules in order to
         enhance extensibility."""
         self.ensure_one()
+        self.refresh()
         if not only_nfp or only_nfp == "out":
             self._calc_qualified_demand()
```
LoisRForgeFlow, line 1223: This method is also used for the nfp calculation, with the refresh in that position it will always be executed, even when you only want to update NFP.

LoisRForgeFlow (changes requested):
Why is it needed? is this solving any bug?

ddmrp/models/stock_buffer.py
```diff
@@ -1218,6 +1218,7 @@ def cron_actions(self, only_nfp=False):
         """This method is meant to be inherited by other modules in order to
         enhance extensibility."""
         self.ensure_one()
+        self.refresh()
```
guewen, line 1221:
```
    @api.model
    def refresh(self):
        """ Clear the records cache.

            .. deprecated:: 8.0
                The record cache is automatically invalidated.
        """
        self.invalidate_cache()
```

The correct method is: `self.invalidate_cache()`

guewen: And actually it looks like a dangerous call, because it wipes all the cache for the stock buffers, I suppose it should be at least `self.invalidate_cache(ids=self.ids)`, but probably `fnames` should be set as well so only the desired computed fields are invalidated.

guewen (changes requested):

ddmrp/models/stock_buffer.py
guewen: I tried something like this:

```
    def cron_actions(self, only_nfp=False):
        """This method is meant to be inherited by other modules in order to
        enhance extensibility."""
        self.ensure_one()
        # invalidate all non-stored computed values
        self.invalidate_cache(
            fnames=[
                fieldname
                for fieldname, props in self._fields.items()
                if props.compute
                and not props.store
                and not props.relational
                and not props.related
            ],
            ids=self.ids,
        )
        if not only_nfp or only_nfp == "out":
```

But even then, it recomputes probably too much, for each buffer of the chunks.

guewen:
> Why is it needed? is this solving any bug?

Same question :)

LoisRForgeFlow:
I think we can close this in favor of the fixes included in https://github.com/OCA/ddmrp/pull/81. @JordiBForgeFlow  If we miss something please reopen it.

PR #61, opened 2020-08-03
[13.0][ddmrp] add method refresh_button in stock.buffer to be called from the buffer views instead of cron_actions

JordiBForgeFlow:


ddmrp/models/stock_buffer.py
```diff
@@ -1214,6 +1214,11 @@ def cron_ddmrp_adu(self, automatic=False):
         _logger.info("End cron_ddmrp_adu.")
         return True
 
+    def refresh_buffer(self):
+        self.ensure_one()
+        self.cron_actions(only_nfp=False)
```
LoisRForgeFlow, line 1219: ensure one is now duplicated in both methods, is is needed in both?

LoisRForgeFlow (changes requested):
Why this change? can you elaborate a little bit?

jgrandguillaume (approved):
LGTM

jbaudoux (approved):

PR #62, opened 2020-08-19
[13.0][IMP] ddmrp: add option to auto procure buffers.

LoisRForgeFlow:
Whenever the buffer is recomputed, if this option is set, it will procure automatically if needed. There are two options to use this feature: standard (when NFP goes below green) or stockout (when NFP goes below 0).

@ForgeFlow

### DIFF
--- a/ddmrp/tests/test_ddmrp.py
+++ b/ddmrp/tests/test_ddmrp.py
@@ -10,88 +10,6 @@ from odoo.addons.ddmrp.tests.common import TestDdmrpCommon
 
 
 class TestDdmrp(TestDdmrpCommon):
-    def create_pickingoutA(self, date_move, qty):
-        return self.pickingModel.with_user(self.user).create(
-            {
-                "picking_type_id": self.ref("stock.picking_type_out"),
-                "location_id": self.binA.id,
-                "location_dest_id": self.customer_location.id,
-                "move_lines": [
-                    (
-                        0,
-                        0,
-                        {
-                            "name": "Test move",
-                            "product_id": self.productA.id,
-                            "date_expected": date_move,
-                            "date": date_move,
-                            "product_uom": self.productA.uom_id.id,
-                            "product_uom_qty": qty,
-                            "location_id": self.binA.id,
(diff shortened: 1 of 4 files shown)


jbaudoux:
@LoisRForgeFlow The procurement should be rather immediate. I remember something was put in place to update the netflow when moves are done. This should also trigger the procurement. The cron is to be used occasionally to update the ADU.

LoisRForgeFlow:
@jbaudoux The auto procurement would happen whenever the buffer is recomputed, so if you have the auto-update NFP option also enabled the procurement will be immediate.

jgrandguillaume (approved):
LGTM thx !

ddmrp/tests/test_ddmrp.py
```diff
@@ -765,6 +765,22 @@ def test_24_purchase_link(self):
...
 
+    def test_25_auto_procure(self):
+        pol = self.pol_model.search([("product_id", "=", self.product_purchased.id)])
+        self.assertFalse(pol)
+        self.assertGreater(self.buffer_purchase.procure_recommended_qty, 0)
+        self.buffer_purchase.auto_procure = True
+        self.buffer_purchase.auto_procure_option = "stockout"
+        self.buffer_purchase.cron_actions()
+        pol = self.pol_model.search([("product_id", "=", self.product_purchased.id)])
+        self.assertFalse(pol)  # Buffer is not in stockout.
```
grindtildeath, line 694: Would be great to have a test covering this case as well.

grindtildeath (approved):
LGTM, just a small non-blocking remark on the test

mileo (approved):
LGTM: Tested on runbot http://3436024-62-92ca31.runbot2.odoo-community.org/web#id=8&action=392&active_id=3&model=purchase.order&view_type=form&cids=&menu_id=101

ddmrp/tests/test_ddmrp.py
LoisRForgeFlow: @grindtildeath Done, I combined it with auto-nfp update for also test both features combined.

PR #66, opened 2020-08-25
[13.0][ddmrp] use split_every in buffer crons in order to allow more frequent commits

JordiBForgeFlow:
...and locking the buffers for too long.

cc @LoisRForgeFlow

### DIFF
--- a/ddmrp/models/stock_buffer.py
+++ b/ddmrp/models/stock_buffer.py
@@ -3,6 +3,7 @@
 
 import logging
 import operator as py_operator
+import threading
 from collections import defaultdict
 from datetime import datetime, timedelta
 from math import pi
@@ -1195,6 +1196,7 @@ class StockBuffer(models.Model):
     @api.model
     def cron_ddmrp_adu(self, automatic=False):
         """calculate ADU for each DDMRP buffer. Called by cronjob."""
+        auto_commit = not getattr(threading.currentThread(), "testing", False)
         _logger.info("Start cron_ddmrp_adu.")
         buffer_ids = self.search([]).ids
         i = 0
@@ -1209,6 +1211,8 @@ class StockBuffer(models.Model):
                             b._calc_adu()
                     else:
                         b._calc_adu()
+                    if auto_commit:
+                        self._cr.commit()
                 except Exception:
                     _logger.exception("Fail to compute ADU for buffer %s", b.name)
                     if not automatic:
@@ -1238,6 +1242,7 @@ class StockBuffer(models.Model):
     def cron_ddmrp(self, automatic=False):
         """Calculate key DDMRP parameters for each buffer.
         Called by cronjob."""
+        auto_commit = not getattr(threading.currentThread(), "testing", False)
         _logger.info("Start cron_ddmrp.")
         buffer_ids = self.search([]).ids
 
@@ -1255,6 +1260,8 @@ class StockBuffer(models.Model):
                     else:
                         b.refresh()
                         b.cron_actions()
+                    if auto_commit:
+                        self._cr.commit()
                 except Exception:
                     _logger.exception("Fail updating buffer %s", b.name)
                     if not automatic:
--- a/ddmrp/models/stock_buffer.py
+++ b/ddmrp/models/stock_buffer.py
@@ -10,6 +10,7 @@ from math import pi
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
 from odoo.tools import float_compare, float_round
+from odoo.tools.misc import split_every
 
 _logger = logging.getLogger(__name__)
 try:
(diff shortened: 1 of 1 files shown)


guewen:
I don't see commits for each chunk, am I missing something?

JordiBForgeFlow:
@guewen you're right. Just added that.

jgrandguillaume (approved):
Thanks, this look good to me !

ddmrp/models/stock_buffer.py
```diff
@@ -1194,23 +1196,27 @@ def write(self, vals):
...
-        j = len(buffers)
-        for b in buffers:
-            try:
-                i += 1
-                _logger.debug("ddmrp cron_adu: {}. ({}/{})".format(b.name, i, j))
-                if automatic:
-                    with self.env.cr.savepoint():
+        j = len(buffer_ids)
+        for buffer_chunk_ids in split_every(50, buffer_ids):
+            for b in self.browse(buffer_chunk_ids):
```
guewen, line 1205:
```suggestion
            for b in self.browse(buffer_chunk_ids).exists():
```
In case a buffer has been deleted.

ddmrp/models/stock_buffer.py
```diff
@@ -1194,23 +1196,27 @@ def write(self, vals):
...
+                    else:
                         b._calc_adu()
-                else:
-                    b._calc_adu()
-            except Exception:
-                _logger.exception("Fail to compute ADU for buffer %s", b.name)
-                if not automatic:
-                    raise
+                    if auto_commit:
+                        self._cr.commit()
```
sebalix, line 1215: Nitpick: Is it not better to put these two lines in a `else` clause of the `try/except` block?

ddmrp/models/stock_buffer.py
```diff
@@ -1236,23 +1242,29 @@ def cron_actions(self, only_nfp=False):
...
+                        b.refresh()
                         b.cron_actions()
-                else:
-                    b.cron_actions()
-            except Exception:
-                _logger.exception("Fail updating buffer %s", b.name)
-                if not automatic:
-                    raise
+                    if auto_commit:
+                        self._cr.commit()
```
sebalix, line 1264: Same remark

ddmrp/models/stock_buffer.py
```diff
@@ -1236,23 +1242,29 @@ def cron_actions(self, only_nfp=False):
...
-        j = len(buffers)
-        buffers.refresh()
-        for b in buffers:
-            i += 1
-            _logger.debug("ddmrp cron: {}. ({}/{})".format(b.name, i, j))
-            try:
-                if automatic:
-                    with self.env.cr.savepoint():
+        j = len(buffer_ids)
+        for buffer_chunk_ids in split_every(50, buffer_ids):
```
guewen, line 1251: @JordiBForgeFlow what was your intention with your chunks? Odoo already prefetches records by chunks of 1000, and you commit every record anyway

ddmrp/models/stock_buffer.py
```diff
@@ -1236,23 +1242,29 @@ def cron_actions(self, only_nfp=False):
...
+            for b in self.browse(buffer_chunk_ids):
+                i += 1
+                _logger.debug("ddmrp cron: {}. ({}/{})".format(b.name, i, j))
+                try:
+                    if automatic:
+                        with self.env.cr.savepoint():
+                            b.refresh()
```
guewen, line 1258: any idea why the cache has to be invalidated on each record? (btw `refresh` is deprecated since 8.0 :wink:

ddmrp/models/stock_buffer.py
guewen: The commit is probably meant to be after one chunk, not after one record, for two reasons:

* the savepoint would be useless if we commit after each record
* in terms of efficiency, a commit after each record is a disaster, as the env is flushed and cleared, so all the records of the chunk are read again (if you have 10'000 buffers, you end up with 10'000 `SELECT` that read 50 ids each time)

ddmrp/models/stock_buffer.py
guewen: @JordiBForgeFlow when doing so, all the ids in buffer_chunk_ids are re-read from database on each loop, so if it's really required, then it's better at least to have the `browse` inside the loop... but that'd be a pity :(

guewen:
I added some comments for issues, but generally, using this brings huge performance issues because for several reasons (commit after each record, `invalidate_cache()` at each iteration, all the records of the chunks are re-read from database on each buffer loop.

ddmrp/models/stock_buffer.py
guewen: Moreover, this PR adds an `invalidate_cache` in `cron_actions` as well: https://github.com/OCA/ddmrp/pull/59

guewen: And same remark as https://github.com/OCA/ddmrp/pull/59#discussion_r568584991, the invalidation of cache should be restricted to some ids and only for the computed fields that needs it, so we don't need to re-read all the buffers on each loop

guewen (changes requested):

guewen:
Added fixes for this in https://github.com/OCA/ddmrp/pull/81

LoisRForgeFlow:
> Added fixes for this in #81

Thanks! Closing in favor of that one.

PR #67, opened 2020-08-27
[13.0][FIX] ddmrp: error when adding a purchase order line

guewen:
To reproduce:

* Open a Purchase Order with at least one line created by a DDMRP buffer
* Add a new line

When selecting the product, we'll get:

```
    File "/odoo/src/odoo/api.py", line 374, in _call_kw_multi
      result = method(recs, *args, **kwargs)
    File "/odoo/src/odoo/models.py", line 6074, in onchange
      snapshot0.fetch(name)

    [...redacted...]

    File "/odoo/src/odoo/models.py", line 3919, in _compute_field_value
      getattr(self, field.compute)()
    File "/odoo/external-src/ddmrp/stock_buffer_capacity_limit/models/stock_buffer.py", line 20, in _compute_procure_recommended_qty
      res = super()._compute_procure_recommended_qty()
    File "/odoo/external-src/ddmrp/ddmrp/models/stock_buffer.py", line 432, in _compute_procure_recommended_qty
      subtract_qty = self.sudo()._quantity_in_progress()
    File "/odoo/external-src/ddmrp/ddmrp/models/stock_buffer.py", line 157, in _quantity_in_progress
      res[buffer.id] += poline.product_uom._compute_quantity(
    KeyError: 3
```

The problem is that, in this context, the field is computed by an
onchange, which wraps the records with a NewID record. The original
record is stored in "_origin".

The idiom:

```
  {}.fromkeys(self.ids, 0.0)
```

Will automatically unwrap the ids, so we prevent comparing a "<NewID
origin=1>" with the actual ID, which is therefore not found.

In _compute_procure_recommended_qty, use _origin if available or skip
the computation.

### DIFF
--- a/ddmrp/models/stock_buffer.py
+++ b/ddmrp/models/stock_buffer.py
@@ -150,7 +150,7 @@ class StockBuffer(models.Model):
     def _quantity_in_progress(self):
         """Return Quantities that are not yet in virtual stock but should
         be deduced from buffers (example: purchases created from buffers)"""
-        res = dict(self.mapped(lambda x: (x.id, 0.0)))
+        res = {}.fromkeys(self.ids, 0.0)
         polines = self.env["purchase.order.line"].search(
             [
                 ("state", "in", ("draft", "sent", "to approve")),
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow:
@guewen I haven't been able to reproduce:

ddmrp-add-pol

maybe you missed a step to reproduce the error?

guewen:
Do you have stock_buffer_capacity_limit installed?

LoisRForgeFlow:
Yes, still no issue

sebalix:
@LoisRForgeFlow Hello, I have been able to reproduce the issue, here is the detailed steps to trigger this `NewId` lookup error:

1. Open a PO form with an existing line having a stock buffer, open the line form, and open the buffer form from there (this will tell the Odoo JS client to fetch these buffer fields the next time we will deal with a PO line)
1. From the PO form, create a new PO line with a product having also a stock buffer (could be the same product than the first line)
1. The onchange will trigger the computation of buffer fields which triggers the above error

This PR could be re-open I guess, it is running on production for months now, working fine.

sebalix (approved):

guewen:
Congrats @sebalix  for finding how to reproduce! 😀

LoisRForgeFlow:
@sebalix Indeed, I've been able to reproduce. That was a sneaky one!!

LoisRForgeFlow (approved):
The fix LGTM as well

sebalix:
Fix ported to 14.0 here: https://github.com/OCA/ddmrp/pull/185

PR #68, opened 2020-09-01
[13.0][FIX] ddmrp: auto update NFP with sudo

LoisRForgeFlow:
When auto updating nfp, action can be triggered by users without write access on buffers, thus we do it with sudo.

@ForgeFlow


mileo (approved):

jgrandguillaume (commented):
LGTM Thx !

TDu (approved):

PR #69, opened 2020-09-02
[13.0][FIX] ddmrp: correct computation of purchased dlt 

LoisRForgeFlow:
correct computation of purchased dlt when there are different lead times for each variant.

@ForgeFlow

### DIFF
--- a/ddmrp/models/stock_buffer.py
+++ b/ddmrp/models/stock_buffer.py
@@ -702,20 +702,30 @@ class StockBuffer(models.Model):
             elif rec.buffer_profile_id.item_type == "distributed":
                 rec.dlt = rec.lead_days
             else:
-                rec.dlt = (
-                    rec.product_id.seller_ids
-                    and rec.product_id.seller_ids[0].delay
-                    or rec.lead_days
-                )
+                sellers = rec._get_product_sellers()
+                rec.dlt = sellers and fields.first(sellers).delay or rec.lead_days
+
+    def _get_product_sellers(self):
+        """:returns the default sellers for a single buffer."""
+        self.ensure_one()
+        all_sellers = self.product_id.seller_ids.filtered(
+            lambda r: not r.company_id or r.company_id == self.company_id
+        )
+        # specific for variant
+        sellers = all_sellers.filtered(lambda s: s.product_id == self.product_id)
+        if not sellers:
+            # generic no variant
+            sellers = all_sellers.filtered(lambda s: not s.product_id)
+        if not sellers:
+            # fallback to all sellers
+            sellers = all_sellers
+        return sellers
 
     @api.depends("buffer_profile_id", "product_id.seller_ids")
     def _compute_main_supplier(self):
         for rec in self:
             if rec.item_type == "purchased":
-                suppliers = rec.product_id.seller_ids.filtered(
-                    lambda r: (not r.product_id or r.product_id == rec.product_id)
-                    and (not r.company_id or r.company_id == rec.company_id)
-                )
+                suppliers = rec._get_product_sellers()
                 rec.main_supplier_id = suppliers[0].name if suppliers else False
             else:
                 rec.main_supplier_id = False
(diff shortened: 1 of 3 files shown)


ddmrp/models/stock_buffer.py
```diff
@@ -702,11 +702,19 @@ def _compute_dlt(self):
...
                 )
+                sellers_no_variant = sellers.filtered(lambda s: not s.product_id)
+                if sellers_variant:
+                    rec.dlt = sellers_variant[0].delay
+                elif sellers_no_variant:
+                    rec.dlt = sellers_no_variant[0].delay
+                elif sellers:
+                    rec.dlt = sellers[0].delay
+                else:
+                    rec.dlt = rec.lead_days
```
simahawk:
```suggestion
                rec.dlt = rec._get_dlt_from_vendor()

    def _get_dlt_from_vendor(self):
        all_sellers = self.product_id.seller_ids
        # specific for variant
        sellers = all_sellers.filtered(
            lambda s: s.product_id == self.product_id
        )
        if not sellers:
            # generic no variant
            sellers = all_sellers.filtered(lambda s: not s.product_id)
        if not sellers:
            # fallback to all sellers
            sellers = all_sellers
        
        if sellers:
            return fields.first(sellers).delay
        return self.lead_days
```

simahawk (changes requested):

ddmrp/models/stock_buffer.py
```diff
@@ -702,11 +702,19 @@ def _compute_dlt(self):
...
-                    rec.product_id.seller_ids
-                    and rec.product_id.seller_ids[0].delay
-                    or rec.lead_days
+                sellers = rec.product_id.seller_ids
+                sellers_variant = sellers.filtered(
+                    lambda s: s.product_id == rec.product_id
                 )
+                sellers_no_variant = sellers.filtered(lambda s: not s.product_id)
+                if sellers_variant:
+                    rec.dlt = sellers_variant[0].delay
```
TDu: If there is multiple suppliers, should they be sorted by delay so it is always the lowest or highest that is selected ?
Same question for the assignations of `rec.dlt` below.

ddmrp/models/stock_buffer.py
LoisRForgeFlow: Don't you think your code is harder to read? My code didn't need any comments as it was self explanatory. Maybe I'm missing something else? I do agree that a dedicated method is always good though

Waiting your thoughts, thanks for the review :)

ddmrp/models/stock_buffer.py
LoisRForgeFlow: That's a good question and if you strictly stick to the theory that's probably true but in reality 99% of the cases the first seller will be used, because in Odoo the first matched seller is used for procurement (thus for the buffer as well) and they are [sorted by sequence by default](https://github.com/odoo/odoo/blob/13.0/addons/product/models/product.py#L711). So, we take the most common use case as the default here.

ddmrp/models/stock_buffer.py
TDu: The solution from @simahawk is better performance wise, as it will filter on some record only if it is necessary.

ddmrp/models/stock_buffer.py
TDu: @LoisRForgeFlow Thank you for your answer.

ddmrp/models/stock_buffer.py
simahawk: exactly this :)

simahawk: in any case the code is quite easy to read even if you don't have specific var names.
What is missing is a good docstring instead. Would you mind adding it?

JordiBForgeFlow (approved):
:+1: technical review. LGTM. @simahawk feedback looks interesting. fields.first? I didn't know that!

ddmrp/models/stock_buffer.py
LoisRForgeFlow: Actually I realized that it is better to have that method to get the default supplier info record and the use it to compute both the main supplier and the dlt.

Would you mind reviewing again?

simahawk (approved):

LoisRForgeFlow:
Last force push just fixed a typo.

/ocabot merge patch

PR #70, opened 2020-09-02
[13.0][FIX] ddmrp_packaging: ensure regular form view is used + remove check_moq method

LoisRForgeFlow:
1st commit
========

when navigating to a package from the buffer. For instance, when
the `delivery` module the delivery view gets precedence.

Before:
Screenshot from 2020-09-02 13-40-21

After:
Screenshot from 2020-09-02 13-40-36

2nd commit
=========
It has been problem to be a problem more than an aid.
As long as the qty_multiple is updated correctly the recommended
procurement will respect it. MOQ field should be reserved to
contractual constraints agreed with the vendor.

@ForgeFlow


jgrandguillaume (approved):
LGTM thx !

JordiBForgeFlow (approved):

jbaudoux (approved):
Perfect. Thanks

PR #71, opened 2020-09-09
[13.0][IMP] ddmrp: automatically assign new MTS PO lines to stock buffers.

LoisRForgeFlow:
If you create a new RFQ or if you duplicate lines in an existing buffer-originated RFQ, it is better to try to find a buffer to link to it.

To do:
* [ ] : Add tests

@ForgeFlow

### DIFF
--- a/ddmrp/models/purchase_order.py
+++ b/ddmrp/models/purchase_order.py
@@ -25,9 +25,24 @@ class PurchaseOrderLine(models.Model):
     ddmrp_comment = fields.Text(related="order_id.ddmrp_comment")
 
     def create(self, vals):
-        record = super(PurchaseOrderLine, self).create(vals)
+        record = super().create(vals)
         record._calc_execution_priority()
+        if record.product_id:
+            record._find_buffer_link()
         return record
+
+    def write(self, vals):
+        res = super().write(vals)
+        for rec in self:
+            if rec.product_id:
+                rec._find_buffer_link()
+        return res
+
+    def _product_id_change(self):
+        res = super()._product_id_change()
+        if self.product_id:
+            self._find_buffer_link()
+        return res
 
     def _calc_execution_priority(self):
         # TODO: handle serveral buffers? worst scenario, average?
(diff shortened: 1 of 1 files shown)


JordiBForgeFlow (approved):
:+1: Functional and technical review. Works as expected.

JordiBForgeFlow:
@didierdonze can you review from your side?

didierdonze:
@JordiBForgeFlow I just checked the code, seems to do what we expect. I ask @guewen for a tech review.

ddmrp/models/purchase_order.py
```diff
@@ -40,3 +55,31 @@ def _calc_execution_priority(self):
...
+            ("product_id", "=", self.product_id.id),
+            ("company_id", "=", self.order_id.company_id.id),
+            ("buffer_profile_id.item_type", "=", "purchased"),
+            ("warehouse_id", "=", self.order_id.picking_type_id.warehouse_id.id),
+        ]
+
+    def _find_buffer_link(self):
+        buffer_model = self.env["stock.buffer"]
+        move_model = self.env["stock.move"]
+        for rec in self.filtered(lambda r: not r.buffer_ids):
```
guewen, line 74: This is looping twice on the records: once to filter the recordset, once in the for loop.
It is more efficient to filter in the same iteration as the for loop:

```
for rec in self:
    if not rec.buffer_ids:
        continue
```

(although in practice, you'll never receive more than one record at a time given the way the method is called now)

ddmrp/models/purchase_order.py
```diff
@@ -25,10 +25,25 @@ class PurchaseOrderLine(models.Model):
...
         record._calc_execution_priority()
+        if record.product_id:
+            record._find_buffer_link()
         return record
 
+    def write(self, vals):
+        res = super().write(vals)
+        for rec in self:
+            if rec.product_id:
+                rec._find_buffer_link()
```
guewen, line 38: instead of adding the same check on product_id everywhere, and calling `_find_buffer_link()`, why not calling `self._find_buffer_link()` and filter in this method at the same time as the filter on `not rec.buffer_ids`?

PR #76, opened 2020-09-24
[13.0] Fix computation for buffer route domain

guewen:
The previous domain for pull rule was comparing the parent location of
the rules' source locations. It may work by chance in some cases, but
doesn't seem accurate.

We are looking for the rule that goes to the buffer, so we have to find
the routes of the rules which destination matches the buffer's location
or one of its parent location.

In fact, it matches the domain of "StockRule._get_rule()". If it
doesn't, the rule would not be applied.

As the rule destination is mandatory, we don't need anymore an explicit
check for the "manufacture" rules.

### DIFF
--- a/stock_buffer_route/models/stock_buffer.py
+++ b/stock_buffer_route/models/stock_buffer.py
@@ -35,18 +35,15 @@ class StockBuffer(models.Model):
 
     def _get_location_routes_of_parents(self, routes, parents):
         return routes.filtered(
-            lambda route: any(
-                p.location_id in parents
-                for p in route.rule_ids.filtered(
-                    lambda rule: rule.action in ("pull", "pull_push", "manufacture")
-                ).mapped("location_src_id")
+            lambda route: (
+                # at least one rule of the route must have a destination location
+                # reaching the buffer
+                route.rule_ids.filtered(lambda rule: rule.action != "push").mapped(
+                    "location_id"
+                )
+                & parents
             )
             or any(rule.action == "buy" for rule in route.rule_ids)
-            or any(
-                rule.action == "manufacture"
-                for rule in route.rule_ids
-                if not rule.location_src_id
-            )
         )
 
     def get_parents(self):


guewen:
Test failure is unrelated...

```
2020-09-24 05:43:29,492 22892 INFO openerp_test odoo.addons.ddmrp.tests.test_ddmrp: ======================================================================

2020-09-24 05:43:29,492 22892 ERROR openerp_test odoo.addons.ddmrp.tests.test_ddmrp: FAIL: TestDdmrp.test_04_adu_calculation_window_past_calendar
Traceback (most recent call last):
File "/home/travis/build/camptocamp/ddmrp/ddmrp/tests/test_ddmrp.py", line 119, in test_04_adu_calculation_window_past_calendar
self.assertAlmostEqual(self.buffer_a.adu, to_assert_value, places=2)
AssertionError: 6.667 != 8.333333333333334 within 2 places
```

sebalix:
Other build are failing due to a similar issue but in another test: https://travis-ci.org/github/OCA/ddmrp/jobs/727057262

stock_buffer_route/models/stock_buffer.py
```diff
@@ -35,18 +35,15 @@ def _compute_route_ids(self):
...
-                for p in route.rule_ids.filtered(
-                    lambda rule: rule.action in ("pull", "pull_push", "manufacture")
-                ).mapped("location_src_id")
+            lambda route: (
+                # at least one rule of the route must have a destination location
+                # reaching the buffer
+                route.rule_ids.filtered(lambda rule: rule.action != "push").mapped(
+                    "location_id"
+                )
+                & parents
```
guewen, line 44: to check: if the rule is `Stock/Shelving/Shelf`, and the buffer's location is `Stock/Shelving`, we are replenishing the buffer, and the reverse is not true. So I don't think we should check the parents of the buffers location, but if the rule destination location is a child of the buffer location.

guewen: My comment is wrong, the current code matches what is done in `_get_rule`.

sebalix (approved):

LoisRForgeFlow (approved):
LGTM :+1: 

Yes, those random errors appear only and occasionally in the travis runs (I have never been able to reproduce locally), most of the times simply restarting the job will fix it - trying now...

LoisRForgeFlow:
Now green! There is probably something some underlying error, but for now we'll stick to the most pragmatic solution as long as it works...
image

/ocabot merge patch

PR #78, opened 2020-10-01
[13.0][IMP] ddmrp: show a warning icon when a buffer has long term supply

LoisRForgeFlow:
Considering "long term supply" what it is not included in the incoming qty used for the NFP calculation.

@ForgeFlow

To do:
* [x] : evaluate performance impact.
* [x] : add tests

### DIFF
--- a/ddmrp/models/stock_buffer.py
+++ b/ddmrp/models/stock_buffer.py
@@ -569,7 +569,9 @@ class StockBuffer(models.Model):
             mrp_data = rec._get_qualified_mrp_moves()
             supply_data = rec._get_incoming_by_days()
             width = timedelta(days=0.4)
-            date_format = self.env["res.lang"]._lang_get(self.env.lang).date_format
+            date_format = (
+                self.env["res.lang"]._lang_get(self.env.lang or "en_US").date_format
+            )
 
             # Plot demand data:
             if demand_data or mrp_data:
--- a/ddmrp/models/stock_buffer.py
+++ b/ddmrp/models/stock_buffer.py
@@ -843,6 +843,15 @@ class StockBuffer(models.Model):
         string="Qualified demand", digits="Product Unit of Measure", readonly=True,
     )
     incoming_dlt_qty = fields.Float(string="Incoming (Within DLT)", readonly=True,)
+    incoming_outside_dlt_qty = fields.Float(
+        string="Incoming (Outside DLT)", readonly=True,
+    )
+    rfq_outside_dlt_qty = fields.Float(
+        string="RFQ Qty (Outside DLT)",
+        readonly=True,
+        help="Request for Quotation total quantity that is planned outside of "
+        "the DLT horizon.",
+    )
     net_flow_position = fields.Float(
         string="Net flow position", digits="Product Unit of Measure", readonly=True,
     )
@@ -1071,9 +1080,8 @@ class StockBuffer(models.Model):
             ("date_expected", "<=", date_to),
         ]
 
-    def _search_stock_moves_incoming_domain(self):
-        self.ensure_one()
-        # We introduce a safety factor of 2 for incoming moves
+    def _get_incoming_supply_date_limit(self):
+        # The safety factor allows to control the date limit
         factor = self.warehouse_id.nfp_incoming_safety_factor or 1
         horizon = int(self.dlt) * factor
         # For purchased products we use calendar days, not work days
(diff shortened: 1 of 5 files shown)


ddmrp/models/stock_buffer.py
```diff
@@ -730,6 +730,59 @@ def _compute_main_supplier(self):
...
+    def _compute_outside_dlt(self):
+        for rec in self:
+            pos = self.env["purchase.order"]
+            mos = self.env["mrp.production"]
+            picks = self.env["stock.picking"]
+            cut_date = rec._get_incoming_supply_date_limit()
+            locations = self.env["stock.location"].search(
+                [("id", "child_of", [rec.location_id.id])]
+            )
+            if rec.item_type == "purchased":
```
jbaudoux: Why isn't it always based on the stock move?
_search_stock_moves_incoming_domain is looking at moves only so why it differs here?
The PO date could differ from the move date if the supplier informs you afterwards of new delivery date

jbaudoux (changes requested):
Computing this on the fly will cause issues.
Why don't you compute this in _calc_incoming_dlt_qty and store the qty inside the horizon and outside the horizon. Having a quantity ouside horizon could also be a good indicator that the buffer must be daily recomputed as the time window is moving each day and at some point in time the qty outside horizon will end-up inside the horizon.

ddmrp/models/stock_buffer.py
jgrandguillaume: So if I understand it well, instead of having a dynamic computation of the warning triangle, we base this one on the outside horizon incoming qty ?
I like this idea ? What do you need ?

LoisRForgeFlow: I agree and I like your suggestions. I'll adapt it

LoisRForgeFlow: There is one issue though and it is RFQ's as they don't yet have stock moves.

LoisRForgeFlow:
@jbaudoux I changed the approach, can you have a look again?

ddmrp/models/stock_buffer.py
```diff
@@ -1256,6 +1284,36 @@ def do_auto_procure(self):
             wizard.make_procurement()
         return True
 
+    def action_view_supply_outside_dlt_window(self):
+        cut_date = self._get_incoming_supply_date_limit()
+        if self.item_type == "purchased":
+            pols = self.purchase_line_ids.filtered(
+                lambda l: l.date_planned > fields.Datetime.to_datetime(cut_date)
+            )
```
jbaudoux, line 1295: I would look at moves outside dlt and take related purchase_line_id because the date_expected of the move could differ from the pol planned date and it is the move date that is considered.

Moreover, in our case, self.purchase_line_ids cannot be used after PO confirmation as we have 1 buffer monitoring a stock location shared by multiple warehouse. This allows us to monitor the global stock. The receipt picking type on the PO can be changed depending on what warehouse you want to send goods to. My concern is here: https://github.com/OCA/ddmrp/blob/13.0/ddmrp/models/purchase_order.py#L68 with the warehouse on the domain. Maybe the link should be created only when the po is created from the buffer as you did for orderpoint_ids? And not recompute each time a write is called on the po line.

Side remark, the m2m link between POL and buffer as stored link will grow each time a PO is created and each time the execution priority is recomputed, it will browse more and more records. Maybe this here https://github.com/OCA/ddmrp/blob/13.0/ddmrp/models/purchase_order.py#L49 needs also to be reviewed a little bit for performance.

Last question. When the buffer is linked to the po line, why the priority computation is launched before the creation of the link? See here: https://github.com/OCA/ddmrp/blob/13.0/ddmrp/models/purchase_order.py#L29

LoisRForgeFlow: 1. Sure, I adapted it to use the move date whenever possible (confirmed PO), otherwise we need to use the date planned in PO lines (RFQ's).
2. This has been a [feature added recently](https://github.com/OCA/ddmrp/pull/71). I tried to cover the most common scenario, but as I knew a perfect fit for everybody was impossible, I include that method as a hook to allow for specific use case customization.
3.  Note that `self`can already be a reduced recordset of PO lines and that that recordset is filtered. As the buffer changes its execution level, you need to update all not closed PO's with the latest value. You should not have too many simultaneous opened purchase orders for a given stock buffer, however I agree that "not closed PO's" sometimes is difficult to define, and the orders remain in `purchase` state even when they are actually shipped and invoiced. We have implemented in some cases a cron job that closes (set status to `done`) PO's already invoiced and delivered.
4. Good catch, I include a commit fixing it in this PR.

jgrandguillaume (approved):
LGTM thx !

jbaudoux (approved):
Thanks

ddmrp/views/stock_buffer_view.xml
```diff
@@ -54,6 +64,13 @@
                     icon="fa-cogs"
                     type="action"
                 />
+                <button
+                    string="See Long Term Supply"
```
JordiBForgeFlow: I would change to "Some incoming quantities are outside of the Decoupled Lead Time Horizon and may require rescheduling. Press this button to display the involved supply orders"

LoisRForgeFlow: Done!

LoisRForgeFlow:
Tests added and button string changed, let's merge this:

/ocabot merge minor

PR #81, opened 2020-10-29
[13.0] Add ddmrp_cron_actions_as_job

guewen:
It makes every calls to "cron_actions" run in queue jobs.

The jobs have an identity key with "identity_exact", meaning that only
one occurence of a job for the same buffer with the same arguments
(only_nfp) will be created at a time (e.g. when the state of a
stock.move is changed several times in the same transaction or in
a different transaction in a short timeframe).

It needs https://github.com/OCA/queue/pull/274 and https://github.com/OCA/queue/pull/275

### DIFF
--- a/ddmrp/models/stock_buffer.py
+++ b/ddmrp/models/stock_buffer.py
@@ -1493,6 +1493,18 @@ class StockBuffer(models.Model):
         """This method is meant to be inherited by other modules in order to
         enhance extensibility."""
         self.ensure_one()
+        self.invalidate_cache(
+            fnames=[
+                "product_location_qty",
+                "incoming_location_qty",
+                "outgoing_location_qty",
+                "virtual_location_qty",
+                "product_location_qty_available_not_res",
+                "dlt",
+                "distributed_source_location_qty",
+            ],
+            ids=self.ids,
+        )
         if not only_nfp or only_nfp == "out":
             self._calc_qualified_demand()
         if not only_nfp or only_nfp == "in":
--- a/ddmrp/models/stock_buffer.py
+++ b/ddmrp/models/stock_buffer.py
@@ -10,6 +10,7 @@ from math import pi
 from odoo import _, api, exceptions, fields, models
 from odoo.exceptions import ValidationError
 from odoo.tools import float_compare, float_round
+from odoo.tools.misc import split_every
 
 _logger = logging.getLogger(__name__)
 try:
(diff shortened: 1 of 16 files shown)


ddmrp/tests/common.py
```diff
@@ -12,6 +12,16 @@ class TestDdmrpCommon(common.SavepointCase):
     def setUpClass(cls):
         super().setUpClass()
 
+        cls.env = cls.env(
+            context=dict(
+                cls.env.context,
+                tracking_disable=True,
```
simahawk, line 18: :tada: :smile:

simahawk (dismissed):
LG

ddmrp_cron_actions_as_job/models/stock_buffer.py
```diff
@@ -0,0 +1,21 @@
+# Copyright 2020 Camptocamp (https://www.camptocamp.com)
+# License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
+
+from odoo import models
+
+from odoo.addons.queue_job.job import identity_exact, job_auto_delay
```
sebalix: No need to protect this import?

guewen: I thought it was no longer required in 13.0? I'll check

guewen: After some tests, I confirm it's no longer needed

sebalix: Didn't know, thanks!

guewen:
https://github.com/OCA/queue/pull/270 has a bug, so I'll propose a temporary alternative solution until I can fix this one

guewen:
I pushed a new implementation that works by patching the method instead of relying on a decorator (see https://github.com/OCA/queue/pull/270#issuecomment-718772150). I still need to work on https://github.com/OCA/queue/pull/274 for this to be ready.

mileo:
Could you add the module to oca_requirements.txt?

guewen:
> Could you add the module to oca_requirements.txt?

done

guewen:
@JordiBForgeFlow @LoisRForgeFlow 

I included the commits of #66 with a new commit fixing issues of #66:

* Call commit once per chunk instead of once per record: commit flushes
  and clears the env, forcing a fetch of all the records of the chunk
  after each line.
* Remove `buffer.refresh()` in cron_ddmrp(): this deprecated method
  calls `invalidate_cache()` without ids and field names, which forces
  a fetch of all the records of the chunk at each line. If it is
  required to invalidate something (but I could not find any
  explanation), we should invalidate only the ids and fields that needs
  invalidation (hopefully only non-stored computed fields). Or at the
  very least, do not fetch the 50 records of the browse at once but one
  by one (but that'd be a pity).

The duration of cron_ddmrp() (when generating jobs with ddmrp_cron_actions_as_job) goes from 8 minutes to 35 seconds.

Pay attention to the usage of `refresh()` / `invalidate_cache()` without specifying what to invalidate. It will force fetching again the records and computations, for all the pre-fetched ids. In the case of `cron_actions`, forcing to read the whole chunk again. As there was no indication why it was necessary, I removed it (when delaying a queue job it's never needed actually).

guewen:
> Pay attention to the usage of `refresh()` / `invalidate_cache()` without specifying what to invalidate. It will force fetching again the records and computations, for all the pre-fetched ids. In the case of `cron_actions`, forcing to read the whole chunk again. As there was no indication why it was necessary, I removed it (when delaying a queue job it's never needed actually).

As some tests failed, I looked again at this and found some fields to invalidate: https://github.com/OCA/ddmrp/pull/81/commits/ce5d553099afab9d04778f29fee987f10a274123, these are the non-stored computed fields used by `cron_actions`

jgrandguillaume (approved):
LGTM for what I can judge about.

ddmrp/models/stock_buffer.py
```diff
@@ -1461,30 +1465,46 @@ def action_view_supply_outside_dlt_window(self):
...
-        j = len(buffers)
-        for b in buffers:
-            try:
-                i += 1
-                _logger.debug("ddmrp cron_adu: {}. ({}/{})".format(b.name, i, j))
-                if automatic:
-                    with self.env.cr.savepoint():
+        j = len(buffer_ids)
+        for buffer_chunk_ids in split_every(self.CRON_DDMRP_CHUNKS, buffer_ids):
+            for b in self.browse(buffer_chunk_ids).exists():
```
LoisRForgeFlow, line 1474: Any particular reason to add these `exists` calls?

LoisRForgeFlow (approved):
LGTM :+1:

Thanks for the detailed commit messages, it eases following the rational behind the changes.

ddmrp/models/stock_buffer.py
LoisRForgeFlow: (in any case this is no blocking and I will continue merging, just willing to know if any problem you encountered forced the change or it was just kind of a good practice)

guewen: We do an initial search to get all the buffer ids, and then we commit: on commit, the db transaction obtains a new snapshot, so if any record has been deleted in another transaction, we'll try to browse it and will fail to read the record when we iterate over it.

LoisRForgeFlow: Got it, thanks.

PR #83, opened 2020-11-04
[13.0] Optimize StockBuffer.cron_actions()

guewen:
Improvements in:

* _calc_qualified_demand
* _calc_incoming_dlt_qty

Before: ~1800ms
After: ~400ms

Tested on a database with 35'000 locations and 32'000 moves.

Details of the profiler below.

Rationale:

In these methods, "*_domain" methods search
all the locations and then filters moves based on their source/destination
locations. In `_calc_incoming_dlt_qty` the search is even done twice.
Searching for the locations first and for the moves with the location
(IN ids, NOT IN ids) domain is inefficient.
Filtering afterwards in memory (based on parent_path comparison) is much
faster and the recordset to filter should not grow too much as
restricted to a single product and a selection of "active" moves.

Also, in case there is no "active" move at all for a buffer, we spare
entirely the cost of searching for the locations (this is a really huge win for the total duration of the cron, on my laptop, computing 4000 buffers goes from 1h30 to 10 minutes).

Profiler details:

**Before**

```
calls     queries   ms
stock.buffer ------------------------ /odoo/external-src/ddmrp/ddmrp/models/stock_buffer.py, 1358

1         0         0.02          @profile
                                  def cron_actions(self, only_nfp=False):
                                      """This method is meant to be inherited by other modules in order to
                                      enhance extensibility."""
1         0         0.02              self.ensure_one()
1         0         0.0               if not only_nfp or only_nfp == "out":
1         24        844.56                self._calc_qualified_demand()
1         0         0.02              if not only_nfp or only_nfp == "in":
1         10        867.96                self._calc_incoming_dlt_qty()
1         6         65.18             self._calc_net_flow_position()
1         0         10.26             self._calc_planning_priority()
1         0         19.66             self._calc_execution_priority()
1         0         0.33              self.mrp_production_ids._calc_execution_priority()
1         1         3.85              self.mapped("purchase_line_ids")._calc_execution_priority()
1         0         0.01              if not only_nfp:
                                          # re-compoute red to force in cascade the recalculation of zones.
1         7         52.63                 self._compute_red_zone()
1         0         0.04              self.do_auto_procure()
1         0         0.0               return True

Total:
1         48        1864.54
```

**After**

```
calls     queries   ms
stock.buffer ------------------------ /odoo/external-src/ddmrp/ddmrp/models/stock_buffer.py, 1351

1         0         0.01          @profile
                                  def cron_actions(self, only_nfp=False):
                                      """This method is meant to be inherited by other modules in order to
                                      enhance extensibility."""
1         0         0.01              self.ensure_one()
1         0         0.0               if not only_nfp or only_nfp == "out":
1         18        162.8                 self._calc_qualified_demand()
1         0         0.01              if not only_nfp or only_nfp == "in":
1         8         63.24                 self._calc_incoming_dlt_qty()
1         9         77.79             self._calc_net_flow_position()
1         0         11.12             self._calc_planning_priority()
1         0         20.84             self._calc_execution_priority()
1         0         0.25              self.mrp_production_ids._calc_execution_priority()
1         1         4.1               self.mapped("purchase_line_ids")._calc_execution_priority()
1         0         0.01              if not only_nfp:
                                          # re-compoute red to force in cascade the recalculation of zones.
1         7         56.11                 self._compute_red_zone()
1         0         0.04              self.do_auto_procure()
1         0         0.01              return True

Total:
1         43        396.33
```

### DIFF
--- a/ddmrp/models/stock_buffer.py
+++ b/ddmrp/models/stock_buffer.py
@@ -1067,9 +1067,6 @@ class StockBuffer(models.Model):
         self.ensure_one()
         horizon = self.order_spike_horizon
         date_to = self.warehouse_id.wh_plan_days(datetime.now(), horizon)
-        locations = self.env["stock.location"].search(
-            [("id", "child_of", [self.location_id.id])]
-        )
         return [
             ("product_id", "=", self.product_id.id),
             (
@@ -1077,10 +1074,17 @@ class StockBuffer(models.Model):
                 "in",
                 ["waiting", "confirmed", "partially_available", "assigned"],
             ),
-            ("location_id", "in", locations.ids),
-            ("location_dest_id", "not in", locations.ids),
             ("date_expected", "<=", date_to),
         ]
+
+    def _search_stock_moves_qualified_demand(self):
+        domain = self._search_stock_moves_qualified_demand_domain()
+        moves = self.env["stock.move"].search(domain)
+        moves = moves.filtered(
+            lambda move: move.location_id.is_sublocation_of(self.location_id)
+            and not move.location_dest_id.is_sublocation_of(self.location_id)
+        )
+        return moves
 
     def _get_incoming_supply_date_limit(self):
         # The safety factor allows to control the date limit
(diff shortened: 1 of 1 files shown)


sebalix (approved):
LG
However I can count already 5 OCA modules sharing the same implementation of `is_sublocation_of` method, we would need a module for that. Maybe a module including a collection of useful methods related to stock? (I'm thinking about some used in `shopfloor`).

simahawk (approved):

guewen:
> However I can count already 5 OCA modules sharing the same implementation of `is_sublocation_of` method, we would need a module for that. Maybe a module including a collection of useful methods related to stock? (I'm thinking about some used in `shopfloor`).

Yes... I had some hope we could get something like this: https://github.com/odoo/odoo/pull/53866
But it won't happen so we should have a base stock module for this.

LoisRForgeFlow:
> > However I can count already 5 OCA modules sharing the same implementation of `is_sublocation_of` method, we would need a module for that. Maybe a module including a collection of useful methods related to stock? (I'm thinking about some used in `shopfloor`).
> 
> Yes... I had some hope we could get something like this: [odoo/odoo#53866](https://github.com/odoo/odoo/pull/53866)
> But it won't happen so we should have a base stock module for this.

Yes, I would be great if they finally consider it, it is indeed very useful.

LoisRForgeFlow (approved):
Thanks for the analysis! changes LGTM

PR #84, opened 2020-11-06
[13.0][FIX] ddmrp: do not consider internal reservations.

LoisRForgeFlow:
Operations done from locations within the buffer area/location should not affect the on hand position of the buffer.

With this change, we rely on `stock.move.line`'s to compute the qty reserved instead of using the `reserved_quantity` on the quant, which is the most accurate way to get the reserved qty. This also eliminates `stock_available_unreserved` dependency.

cc @jbaudoux

@ForgeFlow

### DIFF
--- a/ddmrp/__manifest__.py
+++ b/ddmrp/__manifest__.py
@@ -15,7 +15,6 @@
     "depends": [
         "purchase_stock",
         "mrp_bom_location",
-        "stock_available_unreserved",
         "stock_demand_estimate",
         "web_widget_bokeh_chart",
         "mrp_multi_level",
--- a/ddmrp/models/stock_buffer.py
+++ b/ddmrp/models/stock_buffer.py
@@ -272,9 +272,25 @@ class StockBuffer(models.Model):
         string="Forecast On Location", compute="_compute_product_available_qty"
     )
     product_location_qty_available_not_res = fields.Float(
-        string="Quantity On Location (Unreserved)",
+        string="Quantity On Hand (Unreserved)",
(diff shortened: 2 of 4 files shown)


jbaudoux (approved):
Thanks for the fix.
LGTM

Nevertheless, what is the advantage of computing the available and waiting demand instead of on hands and entire demand? The computation is more complex, so why it is implemented this way?

Any chance you rename the legend here: https://github.com/OCA/ddmrp/blob/13.0/ddmrp/models/stock_buffer.py#L522
with "Available Position" instead of "On-Hand Position" as it is what the figure represents.

LoisRForgeFlow:
@jbaudoux The reason to consider reservations is that the buffers needs to reflect what is available to be consumed, so from a DDMRP perspective the "on-hand" qty is qty available to be consumed right now. If you ignore the reservation you can show a on-hand position of maybe 150% that can lead you to think you are good from an execution perspective, but if all that inventory is reserved, you will find that you cannot complete new orders and therefore have a execution issue. In fact, the net flow equation was initially called "available stock equation".

For the nomenclature issue, it is right that there are some discrepancies between what is on-hand for Odoo and what is for the DDI. I have added some little changes to make everything more explicit.

JordiBForgeFlow (approved):

LoisRForgeFlow:
just squashed commits.

jgrandguillaume (approved):
Thx for the improvements and fix. LGTM

LoisRForgeFlow:
Thanks for the reviews, let's merge it:

/ocabot merge minor

PR #85, opened 2020-11-18
[13.0] [IMP] Add imposed demand lead time to buffers.

JordiBForgeFlow:
When defined, the yellow zone will be computed as the ADU x Imposed Demand Lead Time, if the Imposed Demand Lead
Time is larger than the Replenishment Lead Time. This is particularily useful in situations of infrequent but
periodic demand. E.g. We receive a large order every 30 days, whereas the supplier takes 10 days to supply.

In this case the yellow zone must cover for the entire cycle of 30 days of demand.

In situations with infrequent demand the ADU tends to be very small, and every new order would be treated
as a spike, when in reality this is not an exceptional situation.

image

### DIFF
--- a/ddmrp/models/stock_buffer.py
+++ b/ddmrp/models/stock_buffer.py
@@ -707,17 +707,18 @@ class StockBuffer(models.Model):
             limit=1,
         )
 
-    @api.depends("lead_days", "product_id.seller_ids.delay")
+    @api.depends("lead_days", "product_id.seller_ids.delay", "extra_lead_time")
     def _compute_dlt(self):
         for rec in self:
             if rec.buffer_profile_id.item_type == "manufactured":
                 bom = rec._get_manufactured_bom()
-                rec.dlt = bom.dlt
+                dlt = bom.dlt
             elif rec.buffer_profile_id.item_type == "distributed":
-                rec.dlt = rec.lead_days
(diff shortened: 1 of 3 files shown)


JordiBForgeFlow:
cc @guewen @jgrandguillaume @jbaudoux as discussed.

jbaudoux:
@JordiBForgeFlow Wouldn't it be better to define it as a supplementary lead time that is added to the computed dlt?

That can also be useful for short lead time where you want to increase the safety by imposing an additional coverage. For instance, lead time is 1 day but I always want to have 3 days of stock.

JordiBForgeFlow:
@jbaudoux it's a good idea. I will implement it as you propose.

LoisRForgeFlow:
I would better place it under the section "Buffer size factors" because this new field does that, changes the buffer sizing.

JordiBForgeFlow:
@LoisRForgeFlow @jbaudoux added the field 'Extra lead time' on top of the DLT field.

image

LoisRForgeFlow (approved):
LGTM

jgrandguillaume (approved):
LGTM thx !

LoisRForgeFlow:
Merge conflicts solved.

/ocabot merge minor

PR #86, opened 2020-11-18
[13.0][ddmrp][IMP] create decimal precision 'Average Daily Usage'.

JordiBForgeFlow:
Before the decimal precision assigned to the ADU was the Product Unit of Measure, but sometimes this precision can be set
to no digits, which would drive to an incorrect ADU calculation.

image

### DIFF
--- a/ddmrp/data/decimal_precision_data.xml
+++ b/ddmrp/data/decimal_precision_data.xml
@@ -8,4 +8,12 @@
         <field name="name">Lead Time</field>
         <field name="digits" eval="0" />
     </record>
+    <record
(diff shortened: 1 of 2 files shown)


HviorForgeFlow (approved):

LoisRForgeFlow:
@JordiBForgeFlow you need to install pre-commit and run it

JordiBForgeFlow:
cc @guewen @jgrandguillaume @jbaudoux as discussed.

LoisRForgeFlow (approved):
LGTM

PR #87, opened 2020-11-20
[13.0][IMP] ddmrp: archive buffers when archiving products.

LoisRForgeFlow:
It is clear that when you archive a product you have to archive as well all related buffers, it isn't that clear when you are unarchiving, for now if there is no better proposal we're going to unarchive all buffers as well.

@ForgeFlow

### DIFF
--- a/ddmrp/models/product_product.py
+++ b/ddmrp/models/product_product.py
@@ -10,3 +10,17 @@ class Product(models.Model):
     buffer_ids = fields.One2many(
         comodel_name="stock.buffer", string="Stock Buffers", inverse_name="product_id",
     )
+
+    def write(self, values):
+        res = super().write(values)
+        if "active" in values:
+            buffers = self.env["stock.buffer"].search(
+                [
+                    ("product_id", "in", self.ids),
+                    "|",
+                    ("active", "=", True),
+                    ("active", "=", False),
+                ]
+            )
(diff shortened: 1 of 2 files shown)


jgrandguillaume (approved):
LGTM thx

ddmrp/models/product_product.py
```diff
@@ -10,3 +10,23 @@ class Product(models.Model):
     buffer_ids = fields.One2many(
         comodel_name="stock.buffer", string="Stock Buffers", inverse_name="product_id",
     )
+
+    def action_archive(self):
```
jbaudoux: There are different ways to archive a product. With product variants, removing a variant attribute on the template will archive the corresponding variant without calling action_archive.
This should be done on create/write

jbaudoux (changes requested):
Issue with product variants

ddmrp/models/product_product.py
LoisRForgeFlow: Nice catch, fixed now. Can you re-check?

jbaudoux (approved):
Looks better. Approved

JordiBForgeFlow (approved):

AaronHForgeFlow (approved):

LoisRForgeFlow:
seems stuck, let's try again:

/ocabot merge patch

PR #88, opened 2020-11-24
[13.0][ADD] ddmrp, smart button from picking to buffer

TDu:
Add a smart button on the stock.picking form to open the list of buffer
related to the product in the stock.picking.
By default it is filtered on Procurement Recommended.

### DIFF
--- a/ddmrp/models/stock_picking.py
+++ b/ddmrp/models/stock_picking.py
@@ -12,14 +12,9 @@ class StockPickgin(models.Model):
         self.ensure_one()
         domain = [
             ("product_id", "in", self.mapped("move_lines.product_id.id")),
-            ("warehouse_id", "=", self.picking_type_id.warehouse_id.id),
             ("company_id", "=", self.company_id.id),
         ]
-        return {
-            "type": "ir.actions.act_window",
-            "name": "Stock Buffers",
-            "res_model": "stock.buffer",
-            "view_mode": "tree,form",
-            "domain": domain,
-            "context": {"search_default_procure_recommended": 1},
-        }
+        action = self.env.ref("ddmrp.action_stock_buffer").read()[0]
+        action["domain"] = domain
+        action["context"] = {"search_default_procure_recommended": 1}
+        return action
--- a/ddmrp/models/stock_picking.py
+++ b/ddmrp/models/stock_picking.py
@@ -1,0 +1,20 @@
+# Copyright 2020 Camptocamp SA
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl.html)
+
+from odoo import models
+
+
+class StockPickgin(models.Model):
+    _inherit = "stock.picking"
+
+    def action_stock_buffer_open(self):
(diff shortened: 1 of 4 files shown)


ddmrp/models/stock_picking.py
```diff
@@ -0,0 +1,20 @@
...
+
+    def action_stock_buffer_open(self):
+        """Open a sock.buffer list related to products of the stock.picking."""
+        product_ids = self.mapped("move_lines.product_id.id")
+        return {
+            "type": "ir.actions.act_window",
+            "name": "Stock Buffers",
+            "res_model": "stock.buffer",
+            "view_mode": "tree,form",
+            "domain": [("product_id", "in", product_ids)],
```
LoisRForgeFlow: it is not that easy, there can be more than one buffer per product, this should be implemented similar as how it is done for purchase orders, where each line can be linked to a buffer and its execution estatus is shown for each line in the PO.

TDu: @LoisRForgeFlow Thanks for your comment. Oh, I am not surprised that it could be more complex.
I guess you mean the `_find_buffer_link` function and the `buffer_ids` Many2many on `purchase.order.line`.
Will try to get some inspiration from those.

LoisRForgeFlow: Yes, exactly. Thanks for contributing :)

TDu: I only improved the domain a little and will try to get some more information tomorrow

ddmrp/models/stock_picking.py
```diff
@@ -0,0 +1,25 @@
...
+            ("company_id", "=", self.company_id.id),
+        ]
+        return {
+            "type": "ir.actions.act_window",
+            "name": "Stock Buffers",
+            "res_model": "stock.buffer",
+            "view_mode": "tree,form",
+            "domain": domain,
+            "context": {"search_default_procure_recommended": 1},
+        }
```
sebalix: You could reuse the existing action https://github.com/OCA/ddmrp/blob/13.0/ddmrp/views/stock_buffer_view.xml#L403 (no duplication, translations already done)
```suggestion
        action = self.env.ref("ddmrp.action_stock_buffer").read()[0]
        action["domain"] = domain
        action["context"] = {"search_default_procure_recommended": 1}
        return action
```

ddmrp/models/stock_picking.py
TDu: So the idea of this smart button is not to have access to the exact buffers related to the transfer. But to give a view of all the buffers related to the product in the transfer. As to troubleshoot a situation.
But will also try to implement the `On hand/TOR` on the `stock.move` in the `Operation` tab.

ddmrp/models/stock_picking.py
TDu: @sebalix Thanks, I updated the solution.

jgrandguillaume (approved):
LGTM, tested

jbaudoux (approved):

PR #89, opened 2020-11-28
[13.0][FIX] ddmrp: when ddmrp_auto_update_nfp option is enabled, buffers should also be updated when move location src/dest is changed

jbaudoux:
This fixes compatibility with modules like stock_dynamic_routing in OCA/wms that rewrites the location on pull/push rules.


LoisRForgeFlow (approved):
Code review LGTM. I have restarted travis build because the errors seemed unrelated.

TDu (approved):

PR #90, opened 2020-12-07
[FIX] Do not include inventory loss in ADU computation

sebalix:
Inventory loss moves should not be included in the ADU past/future computation (not relevant).

Ref. 1830


ddmrp/models/stock_buffer.py
```diff
@@ -978,10 +978,14 @@ def _past_demand_estimate_domain(self, date_from, date_to, locations):
...
         self.ensure_one()
+        # Exclude inventory loss moves
+        inventory_locations = self.env["stock.location"].search(
+            [("usage", "=", "inventory")]
+        )
         return [
             ("state", "=", "done"),
             ("location_id", "in", locations.ids),
-            ("location_dest_id", "not in", locations.ids),
+            ("location_dest_id", "not in", (locations | inventory_locations).ids),
```
simahawk: can't we save a search by using `("location_dest_id.usage", "in", (...allowed_values...))`?

sebalix: Good point, but using a dotted syntax in a search domain doesn't generate another query too?

jgrandguillaume (approved):
Functionally speaking, this is ok ! Thx

JordiBForgeFlow (approved):

JordiBForgeFlow:
Makes sense! +1

ddmrp/models/stock_buffer.py
sebalix: After checking the execution query plan of both requests with @guewen , the one with `location_dest_id.usage` could be better at the end with lots of moves, it needs confirmation but I prefer to change it this way for now as it also makes the code clearer about the intent.

guewen (approved):

ddmrp/models/stock_buffer.py
```diff
@@ -982,6 +982,7 @@ def _past_moves_domain(self, date_from, date_to, locations):
             ("state", "=", "done"),
             ("location_id", "in", locations.ids),
             ("location_dest_id", "not in", locations.ids),
+            ("location_dest_id.usage", "!=", "inventory"),
```
LoisRForgeFlow, line 985: Could you include an extra domain based on a company setting to be able to include scraps, something like:
```python
if not self.env.company.ddmrp_adu_calc_include_scrap:
    domain.append(("location_id.scrap_location", "=" , False))
```

cc @jgrandguillaume I have checked the option to use a production location as a scrap location but Odoo only lets you use internal or inventory loss location to scrap.

LoisRForgeFlow:
@sebalix Thanks! LGTM, could you fixup the second commit?

sebalix:
@LoisRForgeFlow done

LoisRForgeFlow (approved):
LGTM, tested locally.

PR #92, opened 2020-12-09
[13.0] Add option to limit replenishment to free quantity 

guewen:
This new feature targets buffers of type 'distributed'.

* Store the source location where the replenishment of a buffer should
  occur.
* The source location is taken from the default first route, so we
  assume the default route is a replenishment, otherwise the computation
  is skipped.
* When `stock_buffer_route` is installed, the route configured on the
  buffer is used to find the replenishment source location.
* A computed field shows the free quantity in the replenishment source
  location.
* When a buffer has no quantity in the replenishment source location, a
  new warning is displayed in the tree view.
* A new option on the buffer profile limits the quantity that can be
  replenished in the procurement wizard to the actual free quantity in
  the replenishment source location. The quantity we can procure for
  this profile will be min(recommended quantity, free quantity).

This is a huge win to avoid backorders for transfers created for
replenishments: we create transfers only for the quantity we have, and
if we do not have this quantity in practice, it reveals a stock issue.

### DIFF
--- a/ddmrp/models/stock_buffer.py
+++ b/ddmrp/models/stock_buffer.py
@@ -1363,13 +1363,37 @@ class StockBuffer(models.Model):
             self._calc_adu()
         return res
 
+    def _procure_qty_to_order(self):
+        qty_to_order = self.procure_recommended_qty
+        rounding = self.procure_uom_id.rounding or self.product_uom.rounding
+        if (
+            self.item_type == "distributed"
+            and self.buffer_profile_id.replenish_distributed_limit_to_free_qty
+        ):
+            # If we don't procure more than what we have in stock, we prevent
+            # backorders on the replenishment
+            if (
+                float_compare(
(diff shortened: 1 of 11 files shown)


sebalix (approved):

TDu (approved):

jbaudoux (changes requested):
Fails with "Quantity must be positive." when the buffer is configured in auto-procure but the free qty in source is 0 as it prevents to create a procurement of quantity 0. This error prevent also the refresh of the buffer (as the refresh triggers the auto-procure that fails).

guewen:
@jbaudoux fixed in my last commit

simahawk (approved):
LG

jbaudoux (approved):

LoisRForgeFlow (approved):
Tested, works smoothly :+1:

PR #94, opened 2020-12-17
[13.0] Add reschedule procurement on distributed buffers

guewen:
Add a configurable time for replenishment procurement on the profile.
This time is added to the planned datetime for the new moves, so they
are considered late when the replenishment time is overpassed.

To be able to do this, I had to change the "date_planned" field on
the procurement wizard from Date to Datetime.

### DIFF
--- a/ddmrp/models/stock_buffer_profile.py
+++ b/ddmrp/models/stock_buffer_profile.py
@@ -28,17 +28,19 @@ class StockBufferProfile(models.Model):
         "variability_id",
         "variability_id.name",
         "variability_id.factor",
+        "distributed_reschedule_max_proc_time",
     )
     def _compute_name(self):
         """Get the right summary for this job."""
         for rec in self:
-            rec.name = "{} {}, {}({}), {}({})".format(
+            rec.name = "{} {}, {}({}), {}({}), {}min".format(
                 rec.replenish_method,
                 rec.item_type,
                 rec.lead_time_id.name,
                 rec.lead_time_id.factor,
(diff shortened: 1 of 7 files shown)


guewen:
@jbaudoux I'm really not confident on this one.
The fact that previously we had a Date in the procurement wizard and now a `Datetime` can be an issue I think. 
But as eventually the `date_planned` field on purchases / `date_expected` field on stock moves is a Datetime as well, I tend to think it was incorrect before... 
It's hard to wrap my head around actually.

TDu (approved):

ddmrp/models/stock_buffer_profile.py
```diff
@@ -62,3 +64,11 @@ def _compute_name(self):
         help="When activated, the recommended quantity will be maxed at "
         "the quantity available in the replenishment source location.",
     )
+    distributed_reschedule_max_proc_time = fields.Float(
```
sebalix, line 67: Should not we put an `onchange` depending on `item_type` to reset `distributed_reschedule_max_proc_time` to 0 as soon as `item_type != 'distributed'`? Or add a check on this where this value is used? If we switch to the type `manufactured` we are still using the existing value, don't know if it's what we want here.

sebalix (commented):
Minor comment.

ddmrp/models/stock_buffer_profile.py
guewen: Added a check where it's used

sebalix (approved):

PR #95, opened 2020-12-18
[13.0][FIX] ddmrp_packaging: propagate contained qty changes

LoisRForgeFlow:
* When the contained quantity of a packaging is updated, it should update the multiple quantity in all the buffers where it is used.
* extra fix in ddmrp: explicitly set the default wizard view.

@ForgeFlow

### DIFF
--- a/ddmrp/wizards/make_procurement_buffer_view.xml
+++ b/ddmrp/wizards/make_procurement_buffer_view.xml
@@ -79,5 +79,6 @@
         <field name="view_mode">form</field>
         <field name="target">new</field>
         <field name="binding_model_id" ref="ddmrp.model_stock_buffer" />
+        <field name="view_id" ref="ddmrp.view_make_procurement_buffer_wizard" />
     </record>
 </odoo>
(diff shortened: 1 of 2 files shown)


jgrandguillaume (changes requested):
Hi @LoisRForgeFlow 

Thanks for support. Look I think this is wrong, if the packaging is upated on the product form, we want to update the quantity multiple (not the packaging multiple).

So here, if the packaging of the product is updated for 32, the multiple quantity should be updated for 64 (not as per now updating the packaging multiple of 2)..Thanks !

Joël

Screenshot from 2020-12-21 08-27-27

jbaudoux (changes requested):
The package multiple should be a fixed multiple of a package. For instance 3 (boxes).
What must change when the quantity of a box is changed is the quantity_multiple.
If a box is 10 units and packaging multiple is 3, then quantity multiple is 30.
If I change the product packaging box to 12 units, then the quantity multiple must automatically change to 36 without having to manually edit the stock buffer

LoisRForgeFlow:
@jgrandguillaume @jbaudoux It has been corrected, please check again.

jbaudoux (approved):
Thanks for the fix

JordiBForgeFlow (approved):

PR #98, opened 2021-02-01
[13.0][FIX] ddmrp: error when clearing produc_id field.

LoisRForgeFlow:
Because compute fields are called before saving, it can
happen that there is no product_id defined (user is removing
it in edit mode). In this situation an error raised, this
commit fixes it.

@ForgeFlow

cc @jgrandguillaume


jgrandguillaume (approved):
LGTM thank you !

marcelsavegnago (approved):
LGTM

PR #104, opened 2021-03-17
[13.0][IMP] ddmrp_sale: able to exclude sale order lines form ADU.

LoisRForgeFlow:
Includes a toggle switch in the sales order lines to exclude related moves from ADU calculation.

image

Pending:
* Be able to exclude quotation lines without moves?
* security, who should have access to this feature? sales manager only?

Extra:
* Should quotation in the past be considered as qualified demand or should they be discarded?

@ForgeFlow


TDu (approved):

jgrandguillaume (approved):
LGTM

PR #107, opened 2021-03-22
[13.0][FIX] ddmrp: main supplier recomputed without respecting the sequence

LoisRForgeFlow:
When the current transaction changed the sequence, it may happen
that the sellers' recordset is not correctly sorted by default.
Doing it explicitly solves the problem.

@ForgeFlow

### DIFF
--- a/ddmrp/models/stock_buffer.py
+++ b/ddmrp/models/stock_buffer.py
@@ -745,6 +745,9 @@ class StockBuffer(models.Model):
         if not sellers:
             # fallback to all sellers
             sellers = all_sellers
+        # When the current transaction changed the sequence, it may happen that
+        # the sellers' recordset is not correctly sorted by default.
+        sellers = sellers.sorted(key="sequence")
         return sellers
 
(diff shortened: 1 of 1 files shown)


JordiBForgeFlow (approved):

bodedra:
@LoisRForgeFlow LG. Just curious to know, Odoo has  `_select_seller` method and in this module has implemented new method `_get_product_sellers`. So what is key different?

LoisRForgeFlow:
> @LoisRForgeFlow LG. Just curious to know, Odoo has `_select_seller` method and in this module has implemented new method `_get_product_sellers`. So what is key different?

@bodedra standard `_select_seller` has extra logic using qty and dates and then sorts by price; this data is not available to compute the main supplier of a buffer (you are not procuring yet and you don't have that information). Also it allows to decouple both logics for custom computations of the main supplier of a buffer.

mileo (approved):

bodedra (approved):
@LoisRForgeFlow It makes sense. Thanks for your clarification.

LG :+1:

PR #108, opened 2021-03-23
[13.0][IMP] ddmrp_warning add applicable domain

TDu:


TDu:
Travis failure seems unrelated ?!

ddmrp_warning/models/ddmrp_warning_definition.py
```diff
@@ -23,6 +24,26 @@ class DdmrpWarningDefinition(models.Model):
         default="mid",
     )
     active = fields.Boolean(default=True)
+    warning_domain = fields.Char(
+        string="Buffer Applicable Domain",
+        default=[],
```
simahawk: a list as default for char?

ddmrp_warning/models/ddmrp_warning_definition.py
```diff
@@ -23,6 +24,26 @@ class DdmrpWarningDefinition(models.Model):
...
+        help="Domain based on Stock Buffer, to define if the "
+        "warning is applicable or not.",
+    )
+
+    def _eval_warning_domain(self, buffer, domain):
+        buffer_domain = [("id", "=", buffer.id)]
+        return bool(
+            self.env["stock.buffer"].search(
```
simahawk: use search_count

simahawk (commented):
LG, small glitches

ddmrp_warning/models/ddmrp_warning_definition.py
TDu: I know strange, but it works as I get the proper value in the db.
I changed it to a string anyway.

ddmrp_warning/models/ddmrp_warning_definition.py
TDu: changed

mileo (approved):

sebalix (approved):

LoisRForgeFlow:
@TDu Why don't you modify the warning definition to consider the filtering you are adding via domain?

I see this domain feature more like a new option, warning definitions can use a python expression OR a domain. I don't see much benefit on combining both.

LoisRForgeFlow (changes requested):
Ongoing discussion.

TDu:
@LoisRForgeFlow Thanks for your review, I actually got that change from a patch file, so can not really answer.
Stay tuned...

jgrandguillaume:
@LoisRForgeFlow combining both is the purpose of this change ;) e.g. in my case I need to apply a certain warning on certain buffer and not other, like checking if the MOQ is bellow multiple qty only apply to certain profile.

Also, it is very nice to segregate certain check on certain type of buffer only. Most likely profile, but not only. I can see may be route_id might also help segragating the warning by buffer kind. 

Applicable domain i a very intuitive way to achieve this IMO.

LoisRForgeFlow:
Hey, sorry for not answering back. I have cancelled the merge, I want to review this a bit more, I will do it today for sure and will merge ASAP.

LoisRForgeFlow:
Ok, let's move this forward. Maybe in the future the performance could be improved by pre-filtering the buffers to run the warning check using the domain. Something like:
```python
...
for w in ddmrp_warnings:
    for buffer in self.env["stock.buffer"].search(w.warning_domain):
        evaluate_definition()
...
```

In anycase not a reason to block it now.

/ocabot merge minor

PR #110, opened 2021-03-26
[13.0][FIX] ddmrp stock.buffer sellers

TDu:
This is to be closer to how it works in Odoo standard.

So supplier wich are specific to the variant or not are
taken into account.


jbaudoux (approved):

jbaudoux:
In std, odoo does not favor sellers with variant, nor exclude the generic one in case you also have one specific to a variant. This change makes it compliant with odoo std behavior

sebalix (approved):

LoisRForgeFlow (approved):
LGTM

PR #112, opened 2021-03-30
[13.0][IMP] ddmrp: add ddmrp_qty_multiple_tolerance option.

LoisRForgeFlow:
Set a tolerance value to apply to quantity multiple in stock
buffers. If the quantity needed is below this tolerance
threshold, the recommended quantity will be reduced a bit
instead of adding another bucket. The value is a percentage
of the quantity multiple.

image

@ForgeFlow


LoisRForgeFlow:
@jbaudoux @jgrandguillaume Could you review?

JordiBForgeFlow (approved):

PR #113, opened 2021-03-31
[13.0][FIX] stock_buffer_capacity_limit: only consider qty on buffered location.

LoisRForgeFlow:
Without this all quantity available for the buffered product was considering, leading to wrong recommendations.

@ForgeFlow

### DIFF
--- a/stock_buffer_capacity_limit/models/stock_buffer.py
+++ b/stock_buffer_capacity_limit/models/stock_buffer.py
@@ -2,6 +2,7 @@
 # License LGPL-3.0 or later (https://www.gnu.org/licenses/lgpl.html).
 
 from odoo import api, fields, models
+from odoo.tools import float_compare
 
 
 class StockBuffer(models.Model):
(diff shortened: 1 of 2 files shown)


jbaudoux:
@LoisRForgeFlow It is missing to take into consideration the qty multiple and min batch. Here is a fix: https://github.com/ForgeFlow/ddmrp-1/pull/1/files

LoisRForgeFlow:
TODO: 
* [x] : Review UoM case.
* [x] : likely depending on this pr https://github.com/OCA/ddmrp/pull/112 to not generate conflitcs because I would like to reuse `_adjust_procure_qty` here.
* [x] : more test cases?

stock_buffer_capacity_limit/models/stock_buffer.py
```diff
@@ -21,10 +22,18 @@ def _compute_procure_recommended_qty(self):
         for rec in self.filtered(lambda b: b.storage_capacity_limit > 0.0):
             recommendation_limit = max(
                 rec.storage_capacity_limit
-                - rec.product_id.qty_available
+                - rec.product_id.with_context(location=rec.location_id.id).qty_available
```
jbaudoux: This is not returning the qty of the location. This solved it:
```suggestion
                - rec.product_location_qty
```
I don't know why, but `rec.product_id.with_context(location=rec.location_id.id).qty_available` does not trigger the computation but reads the result from cache and that contains the qty available of the product, not of the location

LoisRForgeFlow: That's interesting, sometimes the new v13 cache has this surprises for us. I'm commiting your proposal now.

jgrandguillaume:
May be we close that one and merge https://github.com/camptocamp/cosanum_odoo/pull/1314

stock_buffer_capacity_limit/models/stock_buffer.py
```diff
@@ -21,10 +22,18 @@ def _compute_procure_recommended_qty(self):
...
                 rec.storage_capacity_limit
-                - rec.product_id.qty_available
+                - rec.product_location_qty
                 - rec.incoming_dlt_qty,
                 0,
             )
-            qty = min(rec.procure_recommended_qty, recommendation_limit)
-            rec.procure_recommended_qty = qty
+            if self.procure_uom_id:
+                rounding = self.procure_uom_id.rounding
```
simahawk:
```suggestion
                rounding = rec.procure_uom_id.rounding
```
everywhere?

simahawk (changes requested):

LoisRForgeFlow:
> TODO:
> 
> * [x]  : Review UoM case.
> * [x]  : likely depending on this pr #112 to not generate conflitcs because I would like to reuse `_adjust_procure_qty` here.
> * [x]  : more test cases?

All pending tasks done, somo more changes were needed. The storage capcity limit uses the product uom while the qty recommended uses the procure uom, I have adapted to consider this.

I will now squash into commit to keep main branch history simple.

LoisRForgeFlow:
@jbaudoux @jgrandguillaume @simahawk I think this is now more complete and trustful with new tests cases. Can you review on your side?

jgrandguillaume (approved):
LGTM

jbaudoux (approved):

PR #114, opened 2021-03-31
[13.0][FIX] stock_buffer_capacity_limit: only consider qty on buffered location v2

TDu:
This is a temporary duplicate of #113 Just to fix the linting

### DIFF
--- a/stock_buffer_capacity_limit/models/stock_buffer.py
+++ b/stock_buffer_capacity_limit/models/stock_buffer.py
@@ -21,7 +21,7 @@ class StockBuffer(models.Model):
         for rec in self.filtered(lambda b: b.storage_capacity_limit > 0.0):
             recommendation_limit = max(
                 rec.storage_capacity_limit
-                - rec.product_id.qty_available
+                - rec.product_id.with_context(location=rec.location_id.id).qty_available
                 - rec.incoming_dlt_qty,
                 0,
             )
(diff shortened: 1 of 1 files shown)


sebalix (approved):

stock_buffer_capacity_limit/models/stock_buffer.py
```diff
@@ -16,7 +16,12 @@ class StockBuffer(models.Model):
         "execution alerts.",
     )
 
-    @api.depends("storage_capacity_limit")
+    @api.depends(
+        "storage_capacity_limit",
+        "product_location_qty",
```
simahawk: `product_location_qty` AFAIS is not stored, you cannot use it here

sebalix: You are sure? I thought we could, at least to trigger the computation from the UI like would do an onchange. Maybe I'm wrong.

simahawk: normally I get a warning from server logs complaining the the field is not stored and, as such, is useless to hook to its change. Am I missing anything?

TDu: It is not stored indeed.
But I did not know that rule either.
Removed from the depends...

simahawk: bah I cannot find any reference in the code...but https://github.com/odoo/odoo/issues/37468
It seems it happens only if the field has no depends specified?
`_compute_product_available_qty` has no depends....maybe it needs a fix as well... who wins? :smile:

mmequignon: you do

LoisRForgeFlow:
#113 has been merged, I think this is not needed anymore. I'm closing it, please reopen if needed.

PR #117, opened 2021-04-08
[13.0][REF] ddmrp_product_replace: add replace "modes".

LoisRForgeFlow:
Some of the main changes included are:
* replacement can be done to a new buffer, instead of changing
  the product of the replaced buffer. (new mode "new_buffer")
* In the new mode buffers are related via `replaced_by_id` field.
* Boolean `use_replacement_for_buffer_status` allow to consider
  buffers being replaced in NFP (via on-hand and incoming).
* warnings shown for products being replaced when procuring, and
  procure recommendation is now 0 if the buffer is not
  distributed.
* Views adaptations and improvements.
* Old funtionality kept in mode "use_existing".

To Do:
* [x] : Update README.
* [x] : Add tests.

cc @jgrandguillaume @jbaudoux 

@ForgeFlow

### DIFF
--- a/ddmrp_product_replace/models/stock_buffer.py
+++ b/ddmrp_product_replace/models/stock_buffer.py
@@ -8,11 +8,43 @@ from odoo.exceptions import ValidationError
 class StockBuffer(models.Model):
     _inherit = "stock.buffer"
 
+    replaced_by_id = fields.Many2one(
+        comodel_name="stock.buffer",
+        string="Replaced by",
+        help="The product in this buffer is replaced by the product of selected "
+        "buffer. When you replace another buffer:\n - Past Demand of the "
+        "replacement buffer will include the past demand of this product\n"
+        " - Several buffers can be replaced in chained and coexist at the "
+        "same time: A replaces B that replaces C, then A aggregates both B"
+        " and C where B only aggregates C",
+        tracking=True,
+    )
+    replaced_by_alert_text = fields.Char(compute="_compute_replaced_by_alert_text",)
+    replacement_for_ids = fields.One2many(
+        string="Replaces", comodel_name="stock.buffer", inverse_name="replaced_by_id",
+    )
+    replacement_for_count = fields.Integer(compute="_compute_replacement_for_count",)
+    is_replacement_product = fields.Boolean(
+        string="Replacement Product",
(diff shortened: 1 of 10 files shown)


ddmrp_product_replace/models/stock_buffer.py
```diff
@@ -23,6 +55,42 @@ def _check_demand_product_ids(self):
...
+        for rec in self:
+            rec.use_replacement_for_buffer_status = rec.item_type in [
+                "manufactured",
+                "purchased",
+            ]
+
+    def _compute_procure_recommended_qty(self):
+        res = super()._compute_procure_recommended_qty()
+        for rec in self:
+            if rec.replaced_by_id and rec.item_type in ["manufactured", "purchased"]:
```
TDu: `item_type` should be added to `api.depends`

ddmrp_product_replace/models/stock_buffer.py
```diff
@@ -23,6 +55,42 @@ def _check_demand_product_ids(self):
...
+    @api.depends("item_type")
+    def _compute_use_replacement_for_buffer_status(self):
+        for rec in self:
+            rec.use_replacement_for_buffer_status = rec.item_type in [
+                "manufactured",
+                "purchased",
+            ]
+
+    def _compute_procure_recommended_qty(self):
+        res = super()._compute_procure_recommended_qty()
```
TDu: Also to avoid wasted computing time, I think it would be better to  split the record that need to to be computed by the super method and the others

LoisRForgeFlow:
@TDu Your comments have been attended :)

TDu:
Thank you, I saw adding tests is in the todo.
Are you working on it, it seems to be a must have for such big changes

LoisRForgeFlow:
@TDu yes, I will do, I was waiting the functional review feedback before doing it.

jgrandguillaume (commented):
@LoisRForgeFlow Functional test went well ! Everything went well.

Can you see to add the tests so we can merge it ?

Thank you :)

LoisRForgeFlow:
@TDu @jgrandguillaume Test added. Could you update your reviews?

TDu (approved):

PR #118, opened 2021-04-12
[14.0][IMP] ddmrp: add ddmrp_qty_multiple_tolerance option.

LoisRForgeFlow:
Set a tolerance value to apply to quantity multiple in stock
buffers. If the quantity needed is below this tolerance
threshold, the recommended quantity will be reduced a bit
instead of adding another bucket. The value is a percentage
of the quantity multiple.

Forward port of https://github.com/OCA/ddmrp/pull/112.

@ForgeFlow


DavidBForgeFlow (approved):

LoisRForgeFlow:
simple forward port, mergind:

/ocabot merge minor

PR #116, opened 2021-04-07
[13.0] Fix console warnings due to list view renderer

jbaudoux:
The error "Error while parsing JSON SyntaxError: Unexpected token ' in
JSON at position 1" was logged in the console when for example the field
contains a {'no_create': True}. This is not a JSON valid string.
Changed to use the same parsing as in odoo basic_view.

PR #119, opened 2021-04-13
[14.0] ddmrp: Fix console warnings due to list view renderer

LoisRForgeFlow:
The error "Error while parsing JSON SyntaxError: Unexpected token ' in
JSON at position 1" was logged in the console when for example the field
contains a {'no_create': True}. This is not a JSON valid string.
Changed to use the same parsing as in odoo basic_view.

Forward port of https://github.com/OCA/ddmrp/pull/116.

cc @jbaudoux 

@ForgeFlow


AdriaGForgeFlow (approved):

HviorForgeFlow (approved):
LGTM :+1:

PR #122, opened 2021-04-15
[13.0][FIX] main supplier compute

mmequignon:
Updates the `@api.depends` fields on the `_compute_main_supplier` method.

When a `product_supplierinfo` is directly updated (not from the product), the method is never called, and therefore the main_supplier_id isn't updated.

Only the last commit is relevant, others are coming from other opened PRs.

### DIFF
--- a/ddmrp/models/stock_buffer.py
+++ b/ddmrp/models/stock_buffer.py
@@ -745,7 +745,7 @@ class StockBuffer(models.Model):
             sellers = all_sellers
         # When the current transaction changed the sequence, it may happen that
         # the sellers' recordset is not correctly sorted by default.
-        sellers = sellers.sorted(key="sequence")
+        sellers = sellers.sorted(key=lambda s: (s.sequence, -s.min_qty, s.price))
         return sellers
 
     @api.depends(
@@ -756,6 +756,8 @@ class StockBuffer(models.Model):
         "product_id.seller_ids.name",
         "product_id.seller_ids.product_id",
         "product_id.seller_ids.sequence",
+        "product_id.seller_ids.min_qty",
+        "product_id.seller_ids.price",
     )
     def _compute_main_supplier(self):
         for rec in self:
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow:
@mmequignon seems reasonable, can you clean the MR to make it mergeable? Thanks.

ddmrp/models/stock_buffer.py
```diff
@@ -743,12 +743,17 @@ def _get_product_sellers(self):
...
 
-    @api.depends("buffer_profile_id", "product_id.seller_ids")
+    @api.depends(
+        "buffer_profile_id",
+        "item_type",
+        "product_id.seller_ids",
+        "product_id.seller_ids.company_id",
+        "product_id.seller_ids.name",
+        "product_id.seller_ids.product_id",
+        "product_id.seller_ids.sequence",
```
jbaudoux, line 758: The main supplier is in standard based on "sequence, min_qty desc, price"
```suggestion
        "product_id.seller_ids.sequence",
        "product_id.seller_ids.min_qty",
        "product_id.seller_ids.price",
```

ddmrp/models/stock_buffer.py
```diff
@@ -743,12 +743,17 @@ def _get_product_sellers(self):
         if not sellers:
             # fallback to all sellers
             sellers = all_sellers
-        # When the current transaction changed the sequence, it may happen that
-        # the sellers' recordset is not correctly sorted by default.
-        sellers = sellers.sorted(key="sequence")
```
jbaudoux, line 748: According to the comment, then we better do something like:
```
sellers = sellers.sorted(key=lambda s: (s.sequence, -s.min_qty, price))
```
to comply with `_order = 'sequence, min_qty desc, price'`

jbaudoux (changes requested):
Can you also add an help on main_supplier_id with:
The main supplier is the first listed supplier defined on the product that is valid for this product variant. Any date restrictions are not taken into account.

jbaudoux (approved):

sebalix (approved):

LoisRForgeFlow (approved):
LGTM

PR #124, opened 2021-04-21
[13.0][FIX] ddmrp: since date planned is now a datetime, the limit for incoming should also be

LoisRForgeFlow:
Since date planned is now a datetime, the limit for consider incoming moves should also be a datetime and use the
same logic.

Otherwise, a just-created incoming supply could be ignored
when rescheduled to the next day (in UTC, the problem is present
in all timezones but with an offset). This commit includes a
test case to highlight the issue.

cc @jgrandguillaume @jbaudoux 

@ForgeFlow

### DIFF
--- a/ddmrp/models/stock_buffer.py
+++ b/ddmrp/models/stock_buffer.py
@@ -226,10 +226,12 @@ class StockBuffer(models.Model):
 
     # MANUAL PROCUREMENT AND UOM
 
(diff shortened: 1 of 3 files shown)


jgrandguillaume (approved):
LGTM thx you

JordiBForgeFlow (approved):

jbaudoux (approved):

PR #125, opened 2021-04-21
[13.0][FIX] ddmrp: extra lead time should only affect sizing.

LoisRForgeFlow:
Extra lead time is only a sizing factor, the description is
updated to make it clearer and the computation of dlt and
zones adapted.

cc @jgrandguillaume @jbaudoux 

@ForgeFlow


jgrandguillaume (approved):
LGTM Thx you for the change

jbaudoux (approved):

PR #127, opened 2021-04-22
[14.0][FIX] ddmrp: extra lead time should only affect sizing.

LoisRForgeFlow:
Extra lead time is only a sizing factor, the description is
updated to make it clearer and the computation of dlt and
zones adapted.

Forward port of https://github.com/OCA/ddmrp/pull/125.

@ForgeFlow


LoisRForgeFlow:
Keeping v13 and v14 sync'd:

/ocabot merge minor

PR #128, opened 2021-04-26
[13.0][IMP] ddmrp improve po line on hand status

TDu:
Add the quantity ordered
Allow multiple edit
Allow editing scheduled date

### DIFF
--- a/ddmrp/views/purchase_order_line_view.xml
+++ b/ddmrp/views/purchase_order_line_view.xml
@@ -11,6 +11,7 @@
                 create="false"
                 editable="top"
                 delete="false"
+                multi_edit="1"
             >
                 <field name="order_id" widget="many2one" readonly="True" />
                 <field name="partner_id" string="Vendor" />
@@ -19,6 +20,7 @@
                     name="on_hand_percent"
                     options='{"buffer_ids": "buffer_ids", "color_from": "execution_priority_level"}'
                 />
+                <field name="product_qty" readonly="True" />
                 <field name="product_id" readonly="True" />
                 <field
                     name="buffer_ids"
@@ -26,7 +28,7 @@
                     widget="many2many_tags"
                     readonly="1"
                 />
-                <field name="date_planned" widget="date" readonly="True" />
+                <field name="date_planned" widget="date" />
                 <field name="state" />
                 <field name="ddmrp_comment" />
             </tree>
--- a/ddmrp/views/purchase_order_line_view.xml
+++ b/ddmrp/views/purchase_order_line_view.xml
@@ -28,7 +28,7 @@
                     widget="many2many_tags"
                     readonly="1"
                 />
-                <field name="date_planned" widget="date" />
+                <field name="date_planned" />
                 <field name="state" />
                 <field name="ddmrp_comment" />
             </tree>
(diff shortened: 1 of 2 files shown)


jgrandguillaume (approved):
Functional test ok. Very useful change to allow people to move date of reception at line level. Works very nicely with https://github.com/OCA/purchase-workflow/tree/13.0/purchase_delivery_split_date

Business case: Vendore callss to move some position delivery and you need to update the reception date. Using the above module, it work so well and nicely.

LoisRForgeFlow:
This is indeed a good UX improvement!

While I was reviewing I realized the follow-up notes were not editable. I have opened a PR to include the fix here: https://github.com/camptocamp/ddmrp/pull/1.

ddmrp/views/purchase_order_line_view.xml
```diff
@@ -19,14 +20,15 @@
...
+                <field name="product_qty" readonly="True" />
                 <field name="product_id" readonly="True" />
                 <field
                     name="buffer_ids"
                     string="Buffers"
                     widget="many2many_tags"
                     readonly="1"
                 />
-                <field name="date_planned" widget="date" readonly="True" />
+                <field name="date_planned" widget="date" />
```
jbaudoux: Is it timezone safe having widget="date" when non readonly? On the PO editable tree view, there is no widget="date"

simahawk:
@TDu can you check travis? Seems unrelated to your changes tho.

TDu:
I rebased, Travis is :green_apple:

ddmrp/views/purchase_order_line_view.xml
TDu: @jbaudoux I don't know what you mean ? Could you explain ?

mmequignon (approved):

ddmrp/views/purchase_order_line_view.xml
LoisRForgeFlow: I think he means that since the field is a `Datetime`, the fact of using the widget `date` could lead to unexpected results. 

I have tested it and indeed weird thing can happen: if you select say "2021-05-25" you are writing "2021-05-25 00:00 UTC" in the database, and if you are in a "negative" timezone like the US, you will see that after saving the date, it is displayed as "2021-05-24" (conversion to your timezone of the previous UTC datetime).

See below a gif of my testing:
Peek 2021-05-05 11-09

Nice catch @jbaudoux !!

What should we do? Just removing the widget might be the easiest...

ddmrp/views/purchase_order_line_view.xml
```diff
@@ -40,6 +40,11 @@
         <field name="inherit_id" ref="purchase.purchase_order_line_search" />
         <field name="arch" type="xml">
             <filter name="hide_cancelled" position="after">
+                <filter
+                    name="hide_no_buffer"
+                    string="Hide lines without buffer"
```
LoisRForgeFlow: I think just "Buffered" will be a better name for the filter.

TDu: Yes, changed it.

sebalix (approved):

PR #129, opened 2021-04-27
[13.0] Fix security on make procurement on stock buffer

jbaudoux:
Prevent any user to run a procurement. As the procurement is run as super user, no security check was performed and any user could run a procurement.
I added a restriction to users having write access on the buffer.
In our setup, we have different buffer profiles (distributed, purchased) managed by different user groups (resp. stock, purchase). Those have read access to all buffers but can only manage (write) their buffers. So I restrict procurement to only the buffer you manage (this explains why the restriction is on write and not read)

Remove obsolete requested_uid from context

cc @jgrandguillaume @LoisRForgeFlow

### DIFF
--- a/ddmrp/wizards/make_procurement_buffer.py
+++ b/ddmrp/wizards/make_procurement_buffer.py
@@ -106,12 +106,15 @@ class MakeProcurementBuffer(models.TransientModel):
     def make_procurement(self):
         self.ensure_one()
         errors = []
-        # TODO: migrate this feature below
-        # User requesting the procurement is passed by context to be able to
-        # update final MO, PO or trasfer with that information.
-        pg_obj = self.env["procurement.group"].with_context(requested_uid=self.env.user)
+        pg_obj = self.env["procurement.group"]
         procurements = []
         for item in self.item_ids:
+            # As procurement is processed with SUPERUSER_ID (see _run_buy and
+            # _run_manufacture), ensure the current user has write access on
+            # the buffer to make a procurement. Otherwise, any user can request
+            # a procurement on any buffer
+            item.buffer_id.check_access_rights("write")
+            item.buffer_id.check_access_rule("write")
             if item.qty <= 0.0:
                 raise ValidationError(_("Quantity must be positive."))
             if not item.buffer_id:


jgrandguillaume (approved):
For me this checks make sense so +1 

Not everyone able to read buffer should be able to create PO or logistic document. So I think this is usefule addition.

JordiBForgeFlow:
@jbaudoux If you'd want to prevent some users from running procurements, the best approach IMHO would be to limit the access to the action to launch procurements, restricting the action to users in a certain group. What do you think @LoisRForgeFlow ?

jgrandguillaume:
> @jbaudoux If you'd want to prevent some users from running procurements, the best approach IMHO would be to limit the access to the action to launch procurements, restricting the action to users in a certain group. What do you think @LoisRForgeFlow ?

Yes, that's a good idea too

jbaudoux:
> @jbaudoux If you'd want to prevent some users from running procurements, the best approach IMHO would be to limit the access to the action to launch procurements, restricting the action to users in a certain group. What do you think @LoisRForgeFlow ?

The difficulty is that both purchase and stock must have access to the buffers. But stock users should not make purchase procurement (i.e. PO). And purchase users should not make internal warehouse replenishment.

JordiBForgeFlow:
@jbaudoux then perhaps we need an explicit group "Create procurements from stock buffers"?

JordiBForgeFlow:
@jbaudoux ahh, I see what you mean. Different users have separate permissions. Perhaps this is something that you should be able to solve through access rights? People from the warehouse perhaps shall not be able to display purchase buffers?

jbaudoux:
> @jbaudoux ahh, I see what you mean. Different users have separate permissions. Perhaps this is something that you should be able to solve through access rights? People from the warehouse perhaps shall not be able to display purchase buffers?

We tried this but it appeared to be a bad idea because you need the full picture of the process chain to ensure no buffer are missing and the complete routing is there and consistent. So we gave read access to both user groups but restricted create/write/unlink access based on buffer profile (purchase/distributed)

LoisRForgeFlow:
> > @jbaudoux ahh, I see what you mean. Different users have separate permissions. Perhaps this is something that you should be able to solve through access rights? People from the warehouse perhaps shall not be able to display purchase buffers?
> 
> We tried this but it appeared to be a bad idea because you need the full picture of the process chain to ensure no buffer are missing and the complete routing is there and consistent. So we gave read access to both user groups but restricted create/write/unlink access based on buffer profile (purchase/distributed)

This seems a good approach to not overcompliate it. So from my side, I have no problem with the change proposed in this PR, we can always improve it or make it more flexible/customizable in the future if needed.

JordiBForgeFlow:
OK then! Good for me too!

LoisRForgeFlow (approved):

PR #134, opened 2021-05-17
[13.0][FIX] ddmrp_product_replace: "considered as demand" products, also for qualified demand + ability to copy packagings

LoisRForgeFlow:
* This way product replaced by a given buffer can affect the net flow position of the replacing buffer on the demand side as well.

* Ability to copy packagings.

@ForgeFlow

### DIFF
--- a/ddmrp_product_replace/models/stock_buffer.py
+++ b/ddmrp_product_replace/models/stock_buffer.py
@@ -149,6 +149,15 @@ class StockBuffer(models.Model):
             )
         return domain
 
+    def _search_stock_moves_qualified_demand_domain(self):
+        domain = super()._search_stock_moves_qualified_demand_domain()
+        if not self.demand_product_ids:
+            return domain
+        domain = [x for x in domain if x[0] != "product_id"] + [
+            ("product_id", "in", self.demand_product_ids.ids)
+        ]
+        return domain
+
     def action_view_buffers_replaced(self):
         action = self.env.ref("ddmrp.action_stock_buffer")
         result = action.read()[0]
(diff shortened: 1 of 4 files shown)


TDu (approved):

ddmrp_product_replace/models/stock_buffer.py
```diff
@@ -149,6 +149,22 @@ def _search_stock_moves_incoming_domain(self, outside_dlt=False):
...
         return domain
 
+    def _search_stock_moves_qualified_demand_domain(self):
+        domain = super()._search_stock_moves_qualified_demand_domain()
+        if not self.demand_product_ids:
+            return domain
+        index_replace = False
+        for n, clause in enumerate(domain):
+            if isinstance(clause, tuple) and clause[0] == "product_id":
+                index_replace = n
```
simahawk:
```suggestion
                index_replace = n
                break
```

simahawk: AFAIS you are replacing a leaf here.
Normally I do something like this which seems more readable:
```py
domain = [x for x in domain if x[0] != "product_id"] +  [("product_id", "in", self.demand_product_ids.ids)]
```

simahawk (approved):

ddmrp_product_replace/models/stock_buffer.py
```diff
@@ -149,6 +149,22 @@ def _search_stock_moves_incoming_domain(self, outside_dlt=False):
...
+        if isinstance(index_replace, int):
```
mmequignon:
```suggestion
        if index_replace is not False:
```

mmequignon (approved):
technical review only

ddmrp_product_replace/models/stock_buffer.py
LoisRForgeFlow: Yes, I agree. Added :+1:

PR #144, opened 2021-06-08
[13.0][FIX] ddmrp: buffer profiles must be no-update

LoisRForgeFlow:
Any change in the profiles is overridden when updating the module.
Also this trigger the recomputation of the buffer zones, as
they depend on buffer profile changes, this is unwanted.
    
CSV data does not support no-update tag so the date is converted to xml.

cc @jgrandguillaume 

@ForgeFlow

**NOTE: to be merged with nobump.**


JordiBForgeFlow (approved):

jgrandguillaume (approved):
LGTM, thx for the fix

MiquelRForgeFlow:
Travis error is unrelated to this PR, see https://github.com/OCA/ddmrp/issues/145.

LoisRForgeFlow:
Travis error is now solved.

/ocabot merge nobump

PR #146, opened 2021-06-11
[14.0][FIX] ddmrp: buffer profiles must be no-update

LoisRForgeFlow:
Any change in the profiles is overridden when updating the module.
Also this trigger the recomputation of the buffer zones, as
they depend on buffer profile changes, this is unwanted.

CSV data does not support no-update tag so the date is converted to
xml.

Forward port of https://github.com/OCA/ddmrp/pull/144.

@ForgeFlow


MiquelRForgeFlow (approved):

JordiMForgeFlow (approved):

AdriaGForgeFlow (approved):

AaronHForgeFlow (approved):
LGTM

PR #157, opened 2021-07-21
[13.0] Move On-Hand status out of purchase order line

TDu:
In the purchase order form view, the on hand status is visible in the
ordered lines tree view.
This information slows down the load of every large purchase order when
this information is not always usefull.

To improve on this. This change removes the on hand status from the
purchase order form.
And adds a smart button on the from to visualize this information.

### DIFF
--- a/ddmrp/models/purchase_order.py
+++ b/ddmrp/models/purchase_order.py
@@ -11,6 +11,11 @@ class PurchaseOrder(models.Model):
 
     ddmrp_comment = fields.Text(string="Follow-up Notes")
 
+    def action_ddmrp_line_details(self):
+        action = self.env.ref("ddmrp.po_line_execution_action").read()[0]
+        action["domain"] = [("id", "in", self.order_line.ids)]
+        return action
+
 
 class PurchaseOrderLine(models.Model):
     _inherit = "purchase.order.line"
--- a/ddmrp/views/purchase_order_view.xml
+++ b/ddmrp/views/purchase_order_view.xml
@@ -7,17 +7,15 @@
         <field name="model">purchase.order</field>
         <field name="inherit_id" ref="purchase.purchase_order_form" />
         <field name="arch" type="xml">
-            <xpath
-                expr="//field[@name='order_line']//field[@name='date_planned']"
-                position="after"
-            >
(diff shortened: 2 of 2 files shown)


mmequignon (approved):

sebalix (approved):

LoisRForgeFlow:
Maybe a silly question but... how come this field `execution_priority_level` can slow the loading? it is an stored field, I would understand if it was a computed one, but it is not obvious to me on an stored one.

Thanks for clarifying!

LoisRForgeFlow (changes requested):
to mark that I did a question.

TDu:
> Maybe a silly question but... how come this field `execution_priority_level` can slow the loading? it is an stored field, I would understand if it was a computed one, but it is not obvious to me on an stored one.
> 
> Thanks for clarifying!

Hi Lois, thanks for your review, the field that is slowing the load is `on_hand_percent`, the other field `execution_priority_level` was invisible and use to format the value of `on_hand_percent` that is why it was removed as well.

LoisRForgeFlow:
> > Maybe a silly question but... how come this field `execution_priority_level` can slow the loading? it is an stored field, I would understand if it was a computed one, but it is not obvious to me on an stored one.
> > Thanks for clarifying!
> 
> Hi Lois, thanks for your review, the field that is slowing the load is `on_hand_percent`, the other field `execution_priority_level` was invisible and use to format the value of `on_hand_percent` that is why it was removed as well.

Same goes for `on_hand_percent`, it is not a computed field, how does it slow loading?

Sorry for my slow replies here...

TDu:
@LoisRForgeFlow You are correct. Actually there is no computed field on ddmrp `purchase.order.line` changes.
Still this PR seems to have help the customer with slow performance on large purchase order. But I can understand that it would be a shame to remove the buffer information from this view.

LoisRForgeFlow:
Yes, I'm quite doubtful of this move because I think they provide a value. Maybe we can move this change to optional module and control its installation/uninstallation from the ddmrp setting panel?

TDu:
@LoisRForgeFlow  For v14 I took your suggestion and move this change to an optional module. https://github.com/OCA/ddmrp/pull/194

PR #159, opened 2021-07-29
[13.0][ADD] basic stock buffer tree view

TDu:
The default stock.buffer tree view can become slow with a lot of
records.
The idea with this PR is to have a more basic stock.buffer tree view
with less fields that would be a little more responsive.

The existing tree view will be called Stock Buffer Details to
differentiate the two.

Unfortunately it does not seem to be much of a speed improvement, mainly
because of the `_compute_product_available_qty` function which is run
for fields that are too important to remove even in it's basic form.

### DIFF
--- a/ddmrp/views/stock_buffer_view.xml
+++ b/ddmrp/views/stock_buffer_view.xml
@@ -90,6 +90,61 @@
             </tree>
         </field>
     </record>
+    <record id="stock_buffer_basic_view_tree" model="ir.ui.view">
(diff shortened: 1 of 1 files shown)


mmequignon (approved):

sebalix:
This one will be rewritten, we'll keep one view.

sebalix:
To close.

mmequignon:
superseded by https://github.com/OCA/ddmrp/pull/160

PR #161, opened 2021-08-09
[13.0][IMP] ddmrp: add an index on 'stock_move_line.state'

sebalix:
This allows to improve the performance when accessing the list of DDMRP buffers.
The field `_compute_product_available_qty` method of `stock.buffer` is searching for `stock.move.line` records based on their status.

On our customer instance, opening the buffers menu now takes ~1s instead of ~6s. Still I don't know if this change should be included in DDMRP or elsewhere (Odoo core? Will they accept that?)... in the meantime, the slowness spotted occurs only in DDMRP. What is your POV?

### DIFF
--- a/ddmrp/models/__init__.py
+++ b/ddmrp/models/__init__.py
@@ -12,6 +12,7 @@ from . import stock_warehouse
 from . import stock_buffer
 from . import product_template
 from . import stock_move
+from . import stock_move_line
 from . import stock_picking
 from . import res_company
 from . import product_product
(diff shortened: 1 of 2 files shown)


LoisRForgeFlow (commented):
Any effect on operations? does this slow picking processing?

sebalix:
@LoisRForgeFlow we didn't notice slow activities during the transfer creation so far with our current project using it (~4000 transfers per day).

LoisRForgeFlow:
Let's take that as a good test ;)

/ocabot merge minor

PR #170, opened 2021-09-20
[FIX] ddmrp: add bokeh dependency, make pre-commit happy

hailangvn:
This is to explicitly have `bokeh` as external dependency and to make pre-commit green.


LoisRForgeFlow (changes requested):
This is dependency is implicit, see https://github.com/OCA/web/blob/14.0/web_widget_bokeh_chart/__manifest__.py.

No need to put it also here.

PR #175, opened 2021-10-01
[14.0][FIX] ddmrp: fix crash in BOM structure report

alexis-via:


rvalyi (approved):
if one wonder why, here is how the method is called https://github.com/OCA/ddmrp/blob/8a15773db113ffc209c13d9da0f4afb244fe645f/ddmrp/hooks.py#L45

and here is what is done with product_id inside _get_bom:
https://github.com/odoo/odoo/blob/14.0/addons/mrp/report/mrp_report_bom_structure.py#L102

LoisRForgeFlow (approved):
Thanks, probably left during migration

JordiBForgeFlow:
@alexis-via looking into DDMRP? Nice to have you onboard!

PR #180, opened 2021-11-23
[13.0][FIX] Disable unarchiving buffer on product activation

TDu:
Having the stock buffer automatically unarchived when the state of a
product changes is dangerous because it can lead to product being
ordered again when it should not anymore.

The issue encountered seemed to be related to how Odoo handles product
variant. And some changes in the product attribute line of a product
would call a write on the active field of related products variant.

An improvement that could be added (maybe coming soon) is to have on
the product form the information of existing archived stock buffers.

### DIFF
--- a/ddmrp/models/product_product.py
+++ b/ddmrp/models/product_product.py
@@ -13,14 +13,9 @@ class Product(models.Model):
 
     def write(self, values):
         res = super().write(values)
-        if "active" in values:
-            buffers = self.env["stock.buffer"].search(
-                [
-                    ("product_id", "in", self.ids),
-                    "|",
-                    ("active", "=", True),
-                    ("active", "=", False),
-                ]
+        if values.get("active") is False:
+            buffers = (
+                self.env["stock.buffer"].sudo().search([("product_id", "in", self.ids)])
             )
-            buffers.write({"active": values.get("active")})
+            buffers.write({"active": False})
         return res


ddmrp/models/product_product.py
```diff
@@ -13,14 +13,7 @@ class Product(models.Model):
...
-            buffers = self.env["stock.buffer"].search(
-                [
-                    ("product_id", "in", self.ids),
-                    "|",
-                    ("active", "=", True),
-                    ("active", "=", False),
-                ]
-            )
-            buffers.write({"active": values.get("active")})
+        if "active" in values and not values.get("active"):
```
mmequignon:
```suggestion
        if values.get("active") is False:
```
It would be None if the key isn't present in `values`

mmequignon (approved):
Cosmetic remark. LG

LoisRForgeFlow:
@TDu LGTM as a concept, can you attend @mmequignon comment? and also: 

> An improvement that could be added (maybe coming soon) is to have on
the product form the information of existing archived stock buffers.

are you going to this?

ddmrp/models/product_product.py
TDu: Thanks updated

TDu:
@LoisRForgeFlow I added a 2nd commit on the PR for the improvement

sebalix (approved):

TDu:
I just added the sudo

ddmrp/views/product_product_view.xml
```diff
@@ -0,0 +1,36 @@
...
+                    attrs="{'invisible': [('archived_buffer_warning_qty', '=', 0)]}"
+                >
+          This product has <field
+                        name="archived_buffer_warning_qty"
+                    /> archived purchase buffers.
+          <button
+                        type="action"
+                        name="%(archived_buffer_action)d"
+                        string="Check them..."
+                        context="{'product_id': id}"
```
i-vyshnevska:
```suggestion
                        context="{'product_id': active_id}"
```

ddmrp/models/product_product.py
```diff
@@ -11,16 +11,24 @@ class Product(models.Model):
...
                     ("active", "=", False),
                 ]
             )
-            buffers.write({"active": values.get("active")})
+
+    def write(self, values):
+        res = super().write(values)
+        if values.get("active") is False:
+            buffers = self.env["stock.buffer"].search([("product_id", "in", self.ids)])
+            buffers.write({"active": False})
```
i-vyshnevska, line 20: buffer.active field could be related through product_id

TDu:
@LoisRForgeFlow  So I implemented a warning for archived buffers but the customer realised that the search would be much more complicated and they do not want this feature.
For this reason I reverted that commit.
If we really want it in the OCA, I can update the PR.

ddmrp/models/product_product.py
TDu: There can be many buffers for one product and some of them can be archived when the product is still active. So that would not work.

ddmrp/models/product_product.py
```diff
@@ -13,14 +13,7 @@ class Product(models.Model):
...
-                [
-                    ("product_id", "in", self.ids),
-                    "|",
-                    ("active", "=", True),
-                    ("active", "=", False),
-                ]
-            )
-            buffers.write({"active": values.get("active")})
+        if values.get("active") is False:
+            buffers = self.env["stock.buffer"].search([("product_id", "in", self.ids)])
```
simahawk:
```suggestion
            buffers = self.env["stock.buffer"].sudo().search([("product_id", "in", self.ids)])
```
no?

LoisRForgeFlow:
@TDu Hey, it looks good to me. Why did you revert the "add warning..." commit?

TDu:
@LoisRForgeFlow I reverted that commit because when it was tested by the customer they realised that it was not helping and the condition for the warning would need to be much more complicated.

LoisRForgeFlow:
@TDu OK, fair enough. Can you remove the last 2 commits then? as they have no effect, they should not be part of the history.

TDu:
@LoisRForgeFlow  Done

LoisRForgeFlow (approved):

PR #186, opened 2021-12-07
[14.0] Fix quantity_in_progress computation

TDu:
If a purchase order line is related to multiple buffers but not all of
those buffer are included in `self` then the computation crash.


simahawk (approved):

sebalix (approved):

LoisRForgeFlow (approved):
:+1:

PR #191, opened 2022-01-10
[13.0][IMP] ddmrp: add index on 'distributed_source_location_id'

sebalix:
We spotted the following query taking a huge amount of time to execute (~15s) on one of our customer DB:
```sql
SELECT "stock_buffer".id FROM "stock_buffer" WHERE ("stock_buffer"."distributed_source_location_id" IS NOT NULL) ORDER BY "stock_buffer"."id"
```

The filter on `distributed_source_location_qty` is using a search method querying the `distributed_source_location_id` field having no index, so this commit is adding one.

### DIFF
--- a/ddmrp/models/stock_buffer.py
+++ b/ddmrp/models/stock_buffer.py
@@ -1025,6 +1025,7 @@ class StockBuffer(models.Model):
         string="Replenishment Location",
         comodel_name="stock.location",
         readonly=True,
+        index=True,
         help="Source location from where goods will be replenished. "
         "Computed when buffer is refreshed from following the Stock Rules.",
     )


sebalix:
Do not merge, I'm not sure it improves the performance actually.

sebalix:
Query plan without index:
```
 Sort  (cost=1502.77..1517.33 rows=5824 width=4) (actual time=27.494..28.457 rows=5824 loops=1)
   Sort Key: id
   Sort Method: quicksort  Memory: 466kB
   ->  Seq Scan on stock_buffer  (cost=0.00..1138.54 rows=5824 width=4) (actual time=0.035..22.976 rows=5824 loops=1)
         Filter: (distributed_source_location_id IS NOT NULL)
         Rows Removed by Filter: 8430
 Planning Time: 0.241 ms
 Execution Time: 29.206 ms
```
Query plan with an index:
```
Sort  (cost=1486.81..1501.37 rows=5824 width=4) (actual time=5.749..6.241 rows=5824 loops=1)
   Sort Key: id
   Sort Method: quicksort  Memory: 466kB
   ->  Index Scan using stock_buffer_distributed_source_location_id_index on stock_buffer  (cost=0.29..1122.59 rows=5824 width=4) (actual time=0.041..3.460 rows=5824 loops=1)
         Index Cond: (distributed_source_location_id IS NOT NULL)
 Planning Time: 0.124 ms
 Execution Time: 6.621 ms
```
So it still deserves an index.

TDu (approved):

LoisRForgeFlow (approved):

PR #193, opened 2022-01-11
[14.0] Add ddmrp_packaging_product_replace

TDu:
There seems to be a different outcome when doing a product replacement
with packaging set on the related stock buffers between v13 and v14. In
v13 the following test passes.
But on v14 the new stock buffer created has the packaging of the old
product set on them and the constraint `_check_product_packaging` raises
an error.

### DIFF
--- a/ddmrp_packaging_product_replace/readme/DESCRIPTION.rst
+++ b/ddmrp_packaging_product_replace/readme/DESCRIPTION.rst
@@ -1,0 +1,7 @@
+This is a glue module between `ddmrp_packaging` and `ddmrp_product_replace`.
+It disable the copy of the packaging information when using the product
+replacement wizard.
+Without it the new stock buffer created have the packaging of the old product
+set on them and the constraint `_check_product_packaging` raises an error.
+
+Strangely, this is not happening with Odoo 13 ?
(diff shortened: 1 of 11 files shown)


ddmrp_packaging_product_replace/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,7 @@
+This is a glue module between `ddmrp_packaging` and `ddmrp_product_replace`.
+It disable the copy of the packaging information when using the product
+replacement wizard.
+Without it the new stock buffer created have the packaging of the old product
+set on them and the constraint `_check_product_packaging` raises an error.
+
+Strangely, this is not happening with Odoo 13 ?
```
sebalix, line 7: Could it be related to the workaround applied by this v13 commit: https://github.com/OCA/ddmrp/commit/5c0acb3c
that is not required anymore or that requires change to work properly?

TDu: That change has also been merged in 13.0 https://github.com/OCA/ddmrp/blob/e4a60f64c89304afac2353accc57af7d7101ac4f/ddmrp_packaging/models/stock_buffer.py#L18

PR #197, opened 2022-02-07
[13.0] ddmrp: auto_procure when stock increases at replenishment location

jbaudoux:
In auto procure, if the profile is configured in distributed with a limit to free qty, it is possible that not all the required quantity is
replenished. When the quantity increases at the replenishment location, a new auto procurement should be triggered.

cc @sebalix @LoisRForgeFlow @jgrandguillaume


jbaudoux:
ready for review ;) @sebalix @LoisRForgeFlow @jgrandguillaume

sebalix (approved):
LG, would need to be ported to v14 & tested in real conditions

PR #208, opened 2022-02-18
[15.0][FIX] ddmrp: method moved to purchase.order.line model

LoisRForgeFlow:
Forward port of #199


LoisRForgeFlow:
Actually I'm hesitant with this one as the orignal method is still called inside the "_procurement" one (https://github.com/odoo/odoo/blob/15.0/addons/purchase_stock/models/purchase.py#L534). I don't see much inconvenience on applying the buffer to the PO line wherever it comes from.

LoisRForgeFlow:
Clarified in v14 PR, it is good to merge.

/ocabot merge patch

PR #210, opened 2022-02-25
[15.0][FIX] ddmrp: remove mrp_move.py as it is not used

BernatPForgeFlow:


LoisRForgeFlow (changes requested):
Can you just remove all the file? I think it is dead code.

BernatPForgeFlow:
> Can you just remove all the file? I think it is dead code.

@LoisRForgeFlow Done!

LoisRForgeFlow (approved):
Ty!

PR #212, opened 2022-03-10
[15.0][FIX] ddmrp: adu field needs to be explicity written when adu_fixed is set

JordiBForgeFlow:
The onchange only has a visual effect. Actual changes to adu must be made in
the create and write methods.

@LoisRForgeFlow @BernatPForgeFlow


BernatPForgeFlow (approved):
Functional review OK!

LoisRForgeFlow (approved):

PR #226, opened 2022-05-05
[14.0][FIX] save readonly field after onchange is performed

ChrisOForgeFlow:
@ForgeFlow


LoisRForgeFlow:
@ChrisOForgeFlow is this still needed?

ChrisOForgeFlow:
> @ChrisOForgeFlow is this still needed?

I think is necessary because we keep onchange on readonly field, one case, but still necesary

LoisRForgeFlow (approved):

PR #243, opened 2023-03-30
[14.0][IMP] ddmrp: Add hook to get produce delay

BernatPForgeFlow:
A hook is added to get production delay because subcontracting BoMs calculate production time differently, from the time of the first vendor with subcontracted type.


LoisRForgeFlow (approved):
LGTM

PR #246, opened 2023-04-12
[15.0][FIX] ddmrp_*: Adapt call to current thread

BernatPForgeFlow:


LoisRForgeFlow (approved):
Indeed, this is a migration leftover.

LGTM :+1:

PR #253, opened 2023-04-28
[14.0][IMP] ddmrp: Redesign stock buffer form view

BernatPForgeFlow:
- On the top, we will have contextualization about the bufffer.
- On the left side, we will have settings and info about the configuration of the buffer and procurements.
- On the right side, we will have information about the buffer.

Apart from this, you will be able to access the source of ADU, Qualified demand and Incoming (DLT) calculations.

Depends on: https://github.com/OCA/manufacture/pull/1023


LoisRForgeFlow (approved):
It looks much better! A great UX improvement :+1:

PR #266, opened 2023-05-09
[14.0][FIX] ddmrp: Correct function parameters in get stock move values

BernatPForgeFlow:


LoisRForgeFlow (approved):
Nice catch :man_facepalming:

PR #269, opened 2023-05-11
[14.0][FIX] ddmrp: Remove active in domains

BernatPForgeFlow:


LoisRForgeFlow (approved):
LGTM

Just to add more explanation, these active fields do not even exist in standard MO, PO and Picking models

PR #284, opened 2023-06-05
[14.0][FIX] ddmrp: Check sublocations in Stock Buffers

BernatPForgeFlow:

PR #289, opened 2023-06-22
[13.0][FIX] ddmrp: Check sublocations in Stock Buffers

LoisRForgeFlow:
Sometimes, we want to define a Stock Buffer with entire warehouse visibility but we have Stock Demand Estimates or BoMs defined in sublocations. This commit allows to see objects defined in sublocations.

Backport of #284 

@ForgeFlow


LoisRForgeFlow:
standard backport 

/ocabot merge patch

PR #290, opened 2023-06-22
[15.0][FIX] ddmrp: Check sublocations in Stock Buffers

LoisRForgeFlow:
Sometimes, we want to define a Stock Buffer with entire warehouse visibility but we have Stock Demand Estimates or BoMs defined in sublocations. This commit allows to see objects defined in sublocations.

Forward port of #284


LoisRForgeFlow:
standard fwport

/ocabot merge patch

PR #256, opened 2023-05-04
[15.0][FIX] ddmrp, case when origin warehouse has 2 steps to delivery

ChrisOForgeFlow:
FWP: https://github.com/OCA/ddmrp/pull/229

@ForgeFlow

cc: @LoisRForgeFlow

PR #231, opened 2022-07-13
[14.0][IMP][ddmrp] add hooks in make procurement wizard for extensibility

JordiBForgeFlow:
Allow to define the procurement location, name, origin in other modules.

@LoisRForgeFlow @ChrisOForgeFlow  @JuanyDForgeflow

PR #229, opened 2022-06-30
[14.0][FIX] added case when resupply from another warehouse

ChrisOForgeFlow:
@ForgeFlow

PR #213, opened 2022-03-15
[15.0][FIX] ddmrp: get_stock_move_values in stock_rule

BernatPForgeFlow:

PR #299, opened 2023-07-05
[15.0][FW] ddmrp: multiple ports from 14.0

sebalix:
Port of the following PRs from 14.0 to 15.0:

Blacklist of:
- #121 
- #168 
- #189 
- #228: feature removed in #212, nothing to port
- #229: already ported in #256
- #231: already ported in #232
- #226: already included in #262
- #266: already ported in #213

### DIFF
--- a/.oca/oca-port/blacklist/ddmrp.json
+++ b/.oca/oca-port/blacklist/ddmrp.json
@@ -1,0 +1,12 @@
+{
+  "pull_requests": {
+    "121": "(auto) Nothing to port from PR #121",
+    "168": "(auto) Nothing to port from PR #168",
+    "189": "(auto) Nothing to port from PR #189",
+    "228": "Feature removed in #212, nothing to port",
+    "229": "Already ported in #256",
+    "231": "Already ported in #232",
+    "226": "Already included in #262",
+    "266": "Already ported in #213"
+  }
+}


sebalix:
@ChrisOForgeFlow Hello, do you think we need to port the migration script of 3a5646c38f8ca84819cf1f733bfe1d6093d0a6d9 in 15.0?

EDIT: it has been ported in https://github.com/OCA/ddmrp/commit/d6e4a5c4446ee0c74d789d6112a62ace89dbc73f already, I have my answer.

sebalix:
PR #228 should not be ported, this feature has been removed in https://github.com/OCA/ddmrp/pull/212 . I'll update the PR soon.

sebalix:
So everything has been ported with but different commit messages, that's why there were a lot of false-positive.
It remains only PRs to blacklist then.

LoisRForgeFlow (approved):
Thanks for running the analysis :+1: 

PD: I'm glad to see that I keep versions aligned and did not forget any changes in the latest version :smile:

sebalix:
@LoisRForgeFlow small remark, to ease future analysis with `oca-port` it would be better to preserve the commits data thanks to `git cherry-pick`/`git format-patch+git am` (while the content of the commit can be adapted to match the version if needed), that way such commit will be considered as ported :) Here often the commits were fully re-created by hand, that's why there is so much blacklisted PRs.

LoisRForgeFlow:
> @LoisRForgeFlow small remark, to ease future analysis with `oca-port` it would be better to preserve the commits data thanks to `git cherry-pick`/`git format-patch+git am` (while the content of the commit can be adapted to match the version if needed), that way such commit will be considered as ported :) Here often the commits were fully re-created by hand, that's why there is so much blacklisted PRs.

I always use cherry-pick, but I will make sure my colleagues do so too and follow the process even when there is conflicts. Thanks for the heads up!

PR #300, opened 2023-07-05
[14.0][FW] ddmrp: multiple ports from 15.0

sebalix:
Port of the following PRs from 15.0 to 14.0:
- #210
- #212
- #246

Blacklist of:
- #203: nothing to port
- #213: nothing to port
- #219: migration PR
- #232: is a port of #231
- #256: nothing to port


simahawk (approved):
LG

ddmrp/models/stock_buffer.py
```diff
@@ -1917,7 +1917,7 @@ def action_view_future_adu(self):
     @api.model
     def cron_ddmrp_adu(self, automatic=False):
         """calculate ADU for each DDMRP buffer. Called by cronjob."""
-        auto_commit = not getattr(threading.currentThread(), "testing", False)
+        auto_commit = not getattr(threading.current_thread(), "testing", False)
```
LoisRForgeFlow, line 1920: In v14 I can see multiple usages of `threading.current_thread()`. Are you sure this is a needed backport?

LoisRForgeFlow (commented):
Thanks for checking

ddmrp/models/stock_buffer.py
sebalix: Needed, probably not.
I checked on Python 3.7 / 3.8 / 3.9, and both are equivalent:
```python
Python 3.7.3 (default, Oct 31 2022, 14:04:00) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import threading
>>> threading.currentThread == threading.current_thread
True
```
```python
Python 3.8.10 (default, May 26 2023, 14:05:08) 
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import threading
>>> threading.currentThread == threading.current_thread
True
```
```python
Python 3.9.5 (default, Nov 23 2021, 15:27:38) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import threading
>>> threading.current_thread == threading.currentThread
True
```
So I guess it's safe to backport it as it's supported on old Python versions (especially the 3.7 which is the one recommended by Odoo I think for v14? I don't remember).
If you prefer to not backport it you tell me!

LoisRForgeFlow: Fair enough, thanks for double checking! I think we can merge as is.

LoisRForgeFlow (approved):

PR #305, opened 2023-07-11
[13.0][FIX] ddmrp: consider multicompany when finding BoMs

LoisRForgeFlow:
A test case is added to avoid regressions.

PR #306, opened 2023-07-11
[14.0][FIX] ddmrp: consider multicompany when finding BoMs

LoisRForgeFlow:
A test case is added to avoid regressions.

Forward port of #305


LoisRForgeFlow:
Standard fwport

/ocabot merge patch

PR #308, opened 2023-07-31
[14.0][IMP] stock_buffer_route: Add alternative item type in buffer 

BernatPForgeFlow:
Depends on https://github.com/OCA/ddmrp/pull/311

### DIFF
--- a/stock_buffer_route/models/__init__.py
+++ b/stock_buffer_route/models/__init__.py
@@ -1,1 +1,3 @@
+from . import mrp_production
+from . import purchase_order
 from . import stock_buffer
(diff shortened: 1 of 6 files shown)


LoisRForgeFlow:
@BernatPForgeFlow I think here there are several things mixed. Could you keep here only "[IMP] stock_buffer_route: Select route in procurement wizard", and move the other commits to a separate PR (my current thought is that those alternative fields should be either a customization or a separated module) so we can discuss individually?

LoisRForgeFlow:
> @BernatPForgeFlow I think here there are several things mixed. Could you keep here only "[IMP] stock_buffer_route: Select route in procurement wizard", and move the other commits to a separate PR (my current thought is that those alternative fields should be either a customization or a separated module) so we can discuss individually?

Done here https://github.com/OCA/ddmrp/pull/379 and kept this PR for the additional changes.

LoisRForgeFlow:
@BernatPForgeFlow could you now solve conflicts?

PR #310, opened 2023-08-03
[14.0][IMP] ddmrp: Change Manufacturing Orders smart button in buffer

BernatPForgeFlow:
Before: Open only MO procured from the specific buffer.
Now: Open all MO with the same product and warehouse than the buffer.

### DIFF
--- a/ddmrp/migrations/14.0.1.18.4/post-get_manufactured_buffer_links.py
+++ b/ddmrp/migrations/14.0.1.18.4/post-get_manufactured_buffer_links.py
@@ -1,0 +1,10 @@
+from odoo import SUPERUSER_ID, api
+
+
+def migrate(cr, version):
+    env = api.Environment(cr, SUPERUSER_ID, {})
+    mrp_productions = env["mrp.production"].search(
+        [("state", "not in", ["done", "cancel"])]
+    )
+    if mrp_productions:
+        mrp_productions._find_buffer_link()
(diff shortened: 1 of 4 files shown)


LoisRForgeFlow (changes requested):
I think this fix could be even better with the following changes:

* Try to assign a buffer to new MOs automatically even if not created from a stock buffer, as it happens with PO lines already.
* Warehouse is a bit too wide as the domain to find a proper buffer. I would use the finished good location of the MO and check that it is child of the location of the candidate buffer.
* A migration script can be added to try to apply buffer to already existing MOs (maybe only MOs in progress to be a bit conservative?)

BernatPForgeFlow:
> I think this fix could be even better with the following changes:
> 
>     * Try to assign a buffer to new MOs automatically even if not created from a stock buffer, as it happens with PO lines already.
> 
>     * Warehouse is a bit too wide as the domain to find a proper buffer. I would use the finished good location of the MO and check that it is child of the location of the candidate buffer.
> 
>     * A migration script can be added to try to apply buffer to already existing MOs (maybe only MOs in progress to be a bit conservative?)

@LoisRForgeFlow  I like the requested changes and think they are a good improvement.

I just don't see point two though. I think that sometimes the finished good location of the MO is not a sublocation of the buffer. So it wouldn't end up finding which buffer it is. On that, maybe I could add filter by levels on which:
1. Firstly I search by finished good location of the MO
2. If it didn't find a buffer, search by the entire warehouse. 
3. And if neither of both find a buffer, then do not assign any buffer to the MO.

What do you think?

PR #311, opened 2023-08-03
[14.0][IMP] ddmrp: Refactoring of incomings calculation

BernatPForgeFlow:
Refactoring of incomings calculation

1. Reformulate incoming quantities calculation
Split incoming quantities between inside or outside DLT. Make it more heritable, to facilitate consideration of other data sources. Store all the incoming moves and rfqs to always show what was calculated the last time it was refreshed.

2. Add customized Stock Moves view

Other changes: 
- A change is made to the DLT date limit calculation so that refreshing a buffer prior to 12:00:00 UCT, take into account DLT date limit supplies.
- In the calculation of the arrival quantities, the case of receptions in more than one step is taken into account. In these cases there was the possibility of double counting the RFQ and the Stock Move of arrival.

### DIFF
--- a/ddmrp/models/stock_buffer.py
+++ b/ddmrp/models/stock_buffer.py
@@ -6,7 +6,7 @@ import logging
 import operator as py_operator
 import threading
 from collections import defaultdict
-from datetime import datetime, timedelta
+from datetime import datetime, time, timedelta
 from math import pi
 
 from odoo import _, api, exceptions, fields, models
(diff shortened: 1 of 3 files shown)


ddmrp/models/stock_buffer.py
```diff
@@ -1174,6 +1181,7 @@ def _compute_product_vendor_code(self):
         help="Request for Quotation total quantity that is planned outside of "
         "the DLT horizon.",
     )
+    total_outside_dlt_qty = fields.Float(compute="_compute_total_outside_dlt_qty")
```
LoisRForgeFlow: total_incoming_outside_dlt_qty?

LoisRForgeFlow (commented):
Now that you are tackling this... Does this work fine with manufactured items? and if you have manufacturing in 3 steps?

BernatPForgeFlow:
> Now that you are tackling this... Does this work fine with manufactured items? and if you have manufacturing in 3 steps?

@LoisRForgeFlow I have added 4 tests that check if it works for purchased and manufactured items with 1 or 3 step manufacturing.

PR #312, opened 2023-08-03
[14.0][IMP] ddmrp: Filter purchase order lines by its state

BernatPForgeFlow:
Change the filter for purchse order lines from its purchase order state to its own state.
Origin: https://github.com/odoo/odoo/pull/58966


LoisRForgeFlow:
@BernatPForgeFlow Do you mean that this has been fixed in 14.0+ version? Why did this dragged your attention? in other words, is this fixing any current bug?

BernatPForgeFlow:
> @BernatPForgeFlow Do you mean that this has been fixed in 14.0+ version? Why did this dragged your attention? in other words, is this fixing any current bug?

@LoisRForgeFlow No, this is not fixing any bug. I noticed that this could already be fixed in v14 and I think we could put it. Besides, I think it makes sense to put it in case any other module differentiate 'state' at the level of PO Lines.

PR #321, opened 2023-10-19
[16.0][FIX] ddmrp, bom structure report variable lead time should be a boolean in case of empty

ChrisOForgeFlow:
Version 16 raise error bom structure report when lead time is zero, variable should be a boolean not empty string


LoisRForgeFlow (commented):
Hi @ChrisOForgeFlow 

I'm trying to reproduce the error, but I am unable. This is what I get from a simple BoM without LT defined:

image

DLT is displayed fine and LT is a bit weird indeed, but no error raises.

This is the result with your PR applied:

image

LoisRForgeFlow (approved):
After internal discussion, the issue is understood (need other external modules to be reproduced) but the fix is correct in any case.

PR #330, opened 2023-11-09
[13.0][FIX]ddmrp: initialize defaultdict with actual model

GuillemCForgeFlow:
If there is a new model inheriting the actual `stock.buffer` model, such as the `stock.simulation.buffer`, then the method will fail as the values from `self` will be from a different model. This is fixed by changing the initialization to an empty record of self.

cc @ForgeFlow


GuillemCForgeFlow:
@LoisRForgeFlow wdyt?

LoisRForgeFlow (approved):
Sounds good! Thanks for the explanation.

PR #331, opened 2023-11-13
[14.0][FIX]ddmrp: initialize defaultdict with actual model

LoisRForgeFlow:
If there is a new model inheriting the actual `stock.buffer` model, such as the `stock.simulation.buffer`, then the method will fail as the values from `self` will be from a different model. This is fixed by changing the initialization to an empty record of self.

Fwport of #330


LoisRForgeFlow:
rebased after pre-commit issues fix.

/ocabot merge patch

PR #341, opened 2023-11-17
[14.0][IMP] ddmrp*:  Add stored fields and remove non used

BernatPForgeFlow:
All fields will be stored since Stock Buffer is a static entity that refreshes its information via a daily cron or a refresh button.
If we keep compute fields, we can face inconsistencies between the stored data and the computed one.


LoisRForgeFlow (approved):
Functional and code review :+1:

PR #353, opened 2023-12-01
[16.0][FIX] ddmrp: Compute chart fields in list view only when click 

BernatPForgeFlow:


JordiBForgeFlow:
I think that they are not adding much value, with comparison to the slowdown in performance they produce.
So, for now it is a :+1 for me. @LoisRForgeFlow

LoisRForgeFlow (commented):
I think there must be a better compromise, for example, loading the field value only on click of the bubble.

LoisRForgeFlow (approved):
Much cleaner! :1st_place_medal:

PR #355, opened 2023-12-05
[16.0][IMP] stock_buffer_route: Add alternative item type in buffer

BernatPForgeFlow:
Forward port from: https://github.com/OCA/ddmrp/pull/308


LoisRForgeFlow (changes requested):
For future reference: after internal discussion this change was deemed too ad-hoc and therefore is not going to be merged because it adds additional complexity and with a low benefit. Probably better to move to an optional module or to rethink the approach.

PR #361, opened 2023-12-07
[16.0][FIX] ddmrp: rename filter planning_priority_level

JordiBForgeFlow:
Otherwise it throws an error when calling it programatically, as it is confused with the field with the same name


LoisRForgeFlow (approved):
/ocabot merge patch

PR #364, opened 2023-12-13
[14.0][FIX] ddmrp_product_replace: do not copy 'Considered as demand' field

sebalix:
This can leads to a wrong computation of e.g. the qualified demand.

Duplicating a buffer having 'Considered as demand' field set, and changing the product of the new buffer doesn't empty the 'Considered as demand' field, resulting with unrelated products that are taken into account in the computation of the qualified demand.

cc @LoisRForgeFlow


LoisRForgeFlow (approved):
Nice catch :+1:

PR #365, opened 2023-12-14
[16.0][FIX] ddmrp: Reduce the size of demand and supply charts

BernatPForgeFlow:
When we install ddmrp_chatter, the sheet size is reduced and charts are too wide. Reducing width it shows correctly with and without ddmrp_chatter.

Before the fix:
image

After the fix: 
image

### DIFF
--- a/ddmrp/models/stock_buffer.py
+++ b/ddmrp/models/stock_buffer.py
@@ -837,7 +837,7 @@ class StockBuffer(models.Model):
                 y_mrp = list(mrp_data.values())
 
                 p = figure(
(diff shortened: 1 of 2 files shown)


LoisRForgeFlow (changes requested):
I think this is not the best fix, why 450? It works for your screen size, but I assume that just that. The proper fix would make it responsive in the width.

image


You could try to play with `sizing_mode = "stretch_width"`. At the same time we would need to ensure a minimun width and that the demand chart goes below the supply one when the width is too small.

PR #369, opened 2023-12-21
[14.0][REF] ddmrp: Compute dlt depending on buffer location

BernatPForgeFlow:
When we compute DLT for manufactured buffers, we need to consider the location from where we are computing. If the buffered products change between one location and another, the DLTs will be different.

https://github.com/OCA/ddmrp/assets/90243017/0e46692d-4b84-4d9c-a38e-fdab8c5c82dd

Comments:
- Which location should have priority, the location of the BoM or the one passed by context? 
- It would be nice to add a feature in the BoM form that allows to change the context location.

### DIFF
--- a/ddmrp/__manifest__.py
+++ b/ddmrp/__manifest__.py
@@ -14,7 +14,6 @@
     "category": "Warehouse",
     "depends": [
         "purchase_stock",
-        "mrp_bom_location",
         "stock_demand_estimate",
         "web_widget_bokeh_chart",
         "mrp_multi_level",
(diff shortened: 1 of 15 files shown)


LoisRForgeFlow (commented):
This is looking good! I think the next step is to get rid of the mrp_bom_location dependency

BernatPForgeFlow:
> This is looking good! I think the next step is to get rid of the mrp_bom_location dependency

I got rid of the mrp_bom_location dependency. WDYT?

LoisRForgeFlow (approved):
LGTM :+1:

PR #370, opened 2023-12-22
[16.0][FIX] ddmrp: Add in qualified demand only moves with positive quantities

PauBForgeFlow:
Migrated from [PR](https://github.com/OCA/ddmrp/pull/326)


LoisRForgeFlow (changes requested):
Still not clear how to reproduce the issue.

PR #373, opened 2023-12-29
[14.0][FIX] ddmrp: Auto-calculate warehouse when duplicating buffers

BernatPForgeFlow:


LoisRForgeFlow (changes requested):
Can you also take care that the company is correct? The company of the warehouse/location should be used.

Also, could you take the ocasion to skip the tree view when the duplication result is a single buffer and got to the form view directly?

BernatPForgeFlow:
> Can you also take care that the company is correct? The company of the warehouse/location should be used.
> 
> Also, could you take the ocasion to skip the tree view when the duplication result is a single buffer and got to the form view directly?

Done!

LoisRForgeFlow (approved):
Ty!

PR #374, opened 2024-01-09
[13.0][IMP] ddmrp: resize top yellow of execution chart

DavidJForgeFlow:
Resize the top yellow of the execution chart to share half of the top zone with the red zone. This fixes the case when the green zone is big enough to consume entirely the top yellow zone.


LoisRForgeFlow (approved):
Tested, it looks much better

PR #396, opened 2024-02-13
[14.0][FIX] ddmrp: Remove mrp_move from qualified demand calculation

BernatPForgeFlow:
Qualified demand should only consider actual demand, such as outgoing stock moves or any entity that refers to something confirmed and real. MRP Move is an entity similar to Stock Demand Estimate where we forecast future demand and should not be considered as qualified demand.

Stock buffers should consider forecasted future demand via the ADU calculation, sizing the buffer according to that demand. But, then, only at the moment that this demand is confirmed it should be considered as qualified demand.

### DIFF
--- a/ddmrp/models/stock_buffer.py
+++ b/ddmrp/models/stock_buffer.py
@@ -808,7 +808,6 @@ class StockBuffer(models.Model):
 
             # Prepare data:
             demand_data = rec._get_demand_by_days(rec.qualified_demand_stock_move_ids)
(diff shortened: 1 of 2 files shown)


JordiBForgeFlow:
@BernatPForgeFlow I agree.

LoisRForgeFlow (changes requested):
We are still discussing this, I request changes for now to signal that is under discussion

PR #401, opened 2024-02-22
[16.0][IMP] ddmrp_adjustment: give more visibility

DavidJForgeFlow:
When an ADU is affected by a DAF, it will appear a button next to the ADU to check which are being applied and what is the factor.

@ForgeFlow

### DIFF
--- a/ddmrp_adjustment/models/stock_buffer.py
+++ b/ddmrp_adjustment/models/stock_buffer.py
@@ -1,4 +1,4 @@
-# Copyright 2017-20 ForgeFlow S.L. (https://www.forgeflow.com)
+# Copyright 2017-24 ForgeFlow S.L. (https://www.forgeflow.com)
 # License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl.html).
 
 import logging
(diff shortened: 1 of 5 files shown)


ddmrp_adjustment/models/stock_buffer.py
```diff
@@ -22,6 +22,13 @@ class StockBuffer(models.Model):
         "parent buffers.",
     )
 
+    future_daf_chart = fields.Text(
+        string="Future DAF Chart",
+    )
```
LoisRForgeFlow: Remove this

ddmrp_adjustment/models/stock_buffer.py
```diff
@@ -176,3 +184,13 @@ def action_view_demand_to_components(self):
             "view_mode": "tree",
             "domain": [("id", "in", demand_ids)],
         }
+
+    def action_view_affecting_adu(self):
+        demand_ids = (
+            self.env["ddmrp.adjustment"].search(self._daf_to_apply_domain()).ids
```
LoisRForgeFlow: Can you show all future adjustments that apply and will apply? 

You can also add a filter, "current" that is added by default.

But the ability to remove that default filter and see future adjustment is valuable I think.

LoisRForgeFlow (commented):
Some comments below

ddmrp_adjustment/models/stock_buffer.py
DavidJForgeFlow: Done!

PR #409, opened 2024-02-28
[16.0][IMP] ddmrp_exclude_moves_adu_calc_sales

DavidJForgeFlow:
Add tests and also maintain a logic between sale.order.line and stock.move exclude_from_adu. If all the stock moves have the same value, it will propagate to the sale order line.


LoisRForgeFlow (approved):
Nice :+1: 

/ocabot merge patch

PR #427, opened 2024-03-27
[14.0][FIX] ddmrp_adjustment: do not overload original docstring of `_calc_adu`

sebalix:
This has the side-effect of changing the description of the jobs spawned by the `ddmrp_cron_actions_as_job` module:
image

This is misleading for users monitoring the job queue, and generated questions about this change.

cc @LoisRForgeFlow


LoisRForgeFlow (approved):
Indeed, thanks!

/ocabot merge patch

PR #435, opened 2024-04-19
[16.0][IMP] ddmrp_report_part_flow_index: change the way flow index is assigned to buffers

ThiagoMForgeFlow:
@ForgeFlow

### DIFF
--- a/ddmrp_report_part_flow_index/models/stock_buffer.py
+++ b/ddmrp_report_part_flow_index/models/stock_buffer.py
@@ -8,5 +8,43 @@ class StockBuffer(models.Model):
     _inherit = "stock.buffer"
 
     flow_index_group_id = fields.Many2one(
-        "ddmrp.flow.index.group", string="Flow Index Group"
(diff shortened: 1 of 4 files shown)


ddmrp_report_part_flow_index/models/stock_buffer.py
```diff
@@ -8,5 +8,40 @@ class StockBuffer(models.Model):
...
+            frequency_group = flow_index_report[0]["order_frequency_group"]
+            for index_group in flow_index_groups:
+                if index_group.upper_range and index_group.lower_range:
+                    if (
+                        index_group.upper_range
+                        <= frequency_group
+                        <= index_group.lower_range
+                    ):
+                        rec.flow_index_group_id = index_group
+                        return
```
LoisRForgeFlow: This will return the method without assigning a value to all buffers, isn't it? I think you need to use `break` to exit the current for loop and continue with the following buffer ("for rec in self" loop)

LoisRForgeFlow (approved):
LGTM  :+1: 

/ocabot merge minor

PR #442, opened 2024-04-30
[14.0][IMP] ddmrp: Add customized Stock Moves view

BernatPForgeFlow:


LoisRForgeFlow (approved):
Functional and technical review, works as expected :+1:

PR #446, opened 2024-05-06
[16.0][IMP] ddmrp: Add customized Stock Moves view

PauBForgeFlow:

fwport of #442


LoisRForgeFlow (approved):
Functional review

/ocabot merge minor

PR #454, opened 2024-05-14
[15.0][IMP] ddmrp: Add customized Stock Moves view

BernatPForgeFlow:

A stock move can be chained with many other moves and can have two sources, like a dropship. Then, to calculate the source of that stock move, a list of fields ordered by priority is followed, in which if any record is found in any of them, it is returned as a source. For example, if we have a stock move that belongs to a dropship with related sale and purchase, it will always return the sale as source because its the first field in the list. The sources field method is extensible to add any others that are needed.

Source: https://github.com/OCA/ddmrp/pull/442


LoisRForgeFlow (approved):
Functional review

/ocabot merge minor

PR #455, opened 2024-05-14
[16.0][IMP] ddmrp: Make fields invisible or readonly if user not in group

BernatPForgeFlow:
If the user is not in 'Technical / Manage Multiple Stock Locations' group, he won't be able to see the warehouse and location in the procurement wizard.

If the user is not in 'DDMRP / Change quantity in manual procurements from Stock Buffers' group, he won't be able to edit the quantity in the procurement wizard.


LoisRForgeFlow (approved):
Tested, thanks for the fix :+1:

Issue #497, opened 2024-11-19
17.0 branch CI failing


henrybackman:
17.0 branch tests are failing:
[https://github.com/OCA/ddmrp/actions/runs/11780101284/job/33190654835](https://github.com/OCA/ddmrp/actions/runs/11780101284/job/33190654835)

LoisRForgeFlow:
Did you investigate what could have changed that made this start failing?

henrybackman:
I've found that this commit causes the failing test:
- https://github.com/odoo/odoo/commit/1ed6dbeae531dff07a16d50457a16c51db184e40

LoisRForgeFlow:
Hi @henrybackman 

Yes, indeed, we also found that culprit and open a PR/issue to discuss with Odoo https://github.com/odoo/odoo/pull/188846

LoisRForgeFlow:
Fixed with a temporary patch via https://github.com/OCA/ddmrp/pull/500

PR #500, opened 2024-12-02
[17.0][FIX] ddmrp: adapt tests to recent upstream changes.

LoisRForgeFlow:
After https://github.com/odoo/odoo/commit/1ed6dbeae531dff07a16d50457a16c51db184e40 we need to explicitly set routes in the propoer order becouse while in the cache there are not yet sorted by sequence (it requires flush to the database, which we cannot do in a test).


LoisRForgeFlow:
Fixes https://github.com/OCA/ddmrp/issues/497. It is a workaround, while Odoo does not fix the root cause.

PR #503, opened 2025-01-13
[17.0][IMP] ddmrp: change popover css class

ThiagoMForgeFlow:
The inheritance of the popover class was affecting other Odoo views. This way, it only applies to the ddmrp.

@ForgeFlow


LoisRForgeFlow (approved):
:+1:

DavidJForgeFlow (approved):

PR #505, opened 2025-01-31
[17.0][IMP] ddmrp_exclude_moves_adu_calc: run calc adu after toggle exlude_from_adu field

ThiagoMForgeFlow:
To keep the buffer synchronized, the ADU of the affected buffers is forced to recalculate when the exclude_from_adu value of a stock_move is changed.

@ForgeFlow

### DIFF
--- a/ddmrp_exclude_moves_adu_calc/models/stock_move.py
+++ b/ddmrp_exclude_moves_adu_calc/models/stock_move.py
@@ -23,3 +23,10 @@ class StockMove(models.Model):
             )
         for rec in self:
             rec.exclude_from_adu = not rec.exclude_from_adu
+
+    def write(self, vals):
+        res = super().write(vals)
+        if "exclude_from_adu" in vals:
+            out_buffers, in_buffers = self._find_buffers_to_update_nfp()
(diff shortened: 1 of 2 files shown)


ddmrp_exclude_moves_adu_calc/models/stock_move.py
```diff
@@ -23,3 +23,12 @@ def _toggle_exclude_from_adu(self):
             )
         for rec in self:
             rec.exclude_from_adu = not rec.exclude_from_adu
+
+    def write(self, vals):
+        res = super().write(vals)
+        if "exclude_from_adu" in vals:
+            affected_buffers = self.env["stock.buffer"].search(
```
LoisRForgeFlow: you can use `_find_buffers_to_update_nfp` stock.move method to find affected buffers.

DavidJForgeFlow: I changed it to the method you mention and also modified the tests to match the new behavior.

LoisRForgeFlow (approved):
LGTM

PR #506, opened 2025-02-10
[13.0][IMP]ddmrp_chatter: track `procure_recommended_qty` values

GuillemCForgeFlow:
On top of the rest of the Buffer fields to be tracked, as the Procure Recommendation quantity, as it might be interesting to have a traceability of whenever it gets updated

cc @ForgeFlow

### DIFF
--- a/ddmrp_chatter/models/stock_buffer.py
+++ b/ddmrp_chatter/models/stock_buffer.py
@@ -22,3 +22,4 @@ class Buffer(models.Model):
     qty_multiple = fields.Float(tracking=True)
     auto_procure = fields.Boolean(tracking=True)
     auto_procure_option = fields.Selection(tracking=True)
+    procure_recommended_qty = fields.Float(tracking=True)


GuillemCForgeFlow:
ping @LoisRForgeFlow

JordiMForgeFlow (approved):

LoisRForgeFlow (commented):
@GuillemCForgeFlow This field can change a lot of times during the day, I see it more as an operational field rather than a configuration field, so I would not track it in the chatter as it would potentially generate a lot of spam and make you lost the traceability on the configuration changes that are relevant in the chatter (buffer profile, adu method, etc).

If you need to store this operational information, I suggest to you add it to ddmrp_history instead.

GuillemCForgeFlow:
> @GuillemCForgeFlow This field can change a lot of times during the day, I see it more as an operational field rather than a configuration field, so I would not track it in the chatter as it would potentially generate a lot of spam and make you lost the traceability on the configuration changes that are relevant in the chatter (buffer profile, adu method, etc).
> 
> If you need to store this operational information, I suggest to you add it to ddmrp_history instead.

Makes to me as well, will close this then 👍🏿

PR #507, opened 2025-02-21
[18.0][MIG] ddmrp

ThiagoMForgeFlow:
Migration of ddmrp
Depends on 
* https://github.com/OCA/web/pull/3076 
* https://github.com/OCA/stock-logistics-warehouse/pull/2223

@ForgeFlow


LoisRForgeFlow (commented):
The BoM overview throws an error:

image

Could you check?

ThiagoMForgeFlow:
> The BoM overview throws an error:
> 
> image
> 
> Could you check?

Fixed, Thanks!

LoisRForgeFlow (approved):
There are a couple of improvements that we'll need to do in v18 but I think that as a strict migration this is good enough to merge, it will also help advance with the other migration PRs as we unlock the base dependency.

PR #527, opened 2025-03-24
[18.0][MIG]/[ADD] ddmrp_packaging_product_replace

ThiagoMForgeFlow:
Migration of proposed changes in v14.0 that were never merged.
Original PR:
* https://github.com/OCA/ddmrp/pull/193

@ForgeFlow


LoisRForgeFlow:
@ThiagoMForgeFlow Could you "fixup the fixups"? :laughing:

ThiagoMForgeFlow:
> @ThiagoMForgeFlow Could you "fixup the fixups"? 😆

Done!

LoisRForgeFlow:
@ThiagoMForgeFlow Could you also split commit per module changed? I see this PR is modifying also `ddmrp_product_replace`.

ThiagoMForgeFlow:
> @ThiagoMForgeFlow Could you also split commit per module changed? I see this PR is modifying also `ddmrp_product_replace`.

Done, Thanks!

LoisRForgeFlow:
@ThiagoMForgeFlow Could you keep authorship of @TDu in the 2 first commits as they are based on his v14 PR?

ThiagoMForgeFlow:
> @ThiagoMForgeFlow Could you keep authorship of @TDu in the 2 first commits as they are based on his v14 PR?

Authorship added!

ddmrp_packaging_product_replace/readme/DESCRIPTION.md
```diff
@@ -0,0 +1,7 @@
+This is a glue module between ddmrp_packaging and ddmrp_product_replace.
+It disables the copy of the packaging information when using the product
+replacement wizard. Without it the new stock buffer created have the
+packaging of the old product set on them and the constraint
+\_check_product_packaging raises an error.
+
+Strangely, this is not happening with Odoo 13 ?
```
twalter-c2c: I wonder if this line is still worth keeping in v18.

ddmrp_product_replace/wizards/ddmrp_product_replace.py
```diff
@@ -123,15 +123,18 @@ def _do_replacement_use_existing(self):
             "type": "ir.actions.act_window",
         }
 
+    def _get_new_buffer_default_value(self, replaced_buffer):
+        return dict(
+            product_id=self.new_product_id.id,
```
twalter-c2c, line 128: Thanks for improving the readability by moving this part to a separate method. You should use the `replaced_buffer` instead of `self` if I am not mistaken.
```suggestion
            product_id=replaced_buffer.new_product_id.id,
```

twalter-c2c (commented):
Thanks! Looks good, small remarks to check.

ddmrp_packaging_product_replace/readme/DESCRIPTION.md
ThiagoMForgeFlow: You are right, line removed.

ddmrp_product_replace/wizards/ddmrp_product_replace.py
ThiagoMForgeFlow: We should use `self` because the field `new_product_id` is from `ddmrp.product.replace` model, not `stock.buffer`. But I have removed the `replaced_buffer` parameter because is not needed in the function.

twalter-c2c: You are right, thank you for the explanation. So yes, the `replaced_buffer` was not needed indeed (and even misleading apparently) in such a case.

twalter-c2c (approved):

LoisRForgeFlow (approved):
:+1:

PR #528, opened 2025-03-31
[18.0][IMP] ddmrp: consider that done quantity and demand can be different in stock.move

LoisRForgeFlow:
Since v17, `quantity` field represents the actual done quantity and
can differ from `product_qty` which remeains as the original demand
when validating a transfer without a backorder. This need to be considered by ddmrp.

To do:

- [x] : Create a generic module that computes stock.move's `quantity` in the uom of the product.
- [x] : Use it here to fix the provided test case.

### DIFF
--- a/ddmrp/README.rst
+++ b/ddmrp/README.rst
@@ -72,12 +72,12 @@ a wide variety of leading companies across the world.
 
 Some of the benefits reported by the DDMRP method include:
 
-- High fill rate performance
-- Lead time reductions
-- Inventory reductions, while improving customer service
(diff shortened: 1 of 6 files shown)


ThiagoMForgeFlow:
Generic module that computes stock.move's quantity in the uom of the product:
* https://github.com/OCA/stock-logistics-workflow/pull/1915

DavidJForgeFlow (approved):
LGTM! Code review :+1:

ThiagoMForgeFlow (approved):

PR #531, opened 2025-05-05
[18.0][FIX] ddmrp: set unique product default codes for tests

henrybackman:
`product-code-unique` module in `product-attribute`  adds a unique constraint to the product codes and when running tests with both modules installed, the test data creation fails in `common.py`


henrybackman:
Replaced by:
- https://github.com/OCA/ddmrp/pull/532

PR #532, opened 2025-05-05
[18.0][FIX] ddmrp: set unique product default codes for tests

henrybackman:
`product-code-unique` module in `product-attribute` adds a unique constraint to the product codes and when running tests with both modules installed, the test data creation fails in `common.py`
- https://github.com/OCA/product-attribute/blob/18.0/product_code_unique/models/product.py


LoisRForgeFlow (approved):
Fair enough

/ocabot merge patch

PR #533, opened 2025-05-07
[18.0][FIX] ddmrp: fix tests with unique default code (again)

henrybackman:
Previous commit changed one code but I forgot to check if it was the only one causing issues :man_facepalming: 
Previous PR
- https://github.com/OCA/ddmrp/pull/532

When unique constraint is set on product.product default_code the tests fail due to a duplicate default_code

### DIFF
--- a/ddmrp/tests/common.py
+++ b/ddmrp/tests/common.py
@@ -217,7 +217,7 @@ class TestDdmrpCommon(common.TransactionCase):
                 "name": "Product C",
                 "is_storable": True,
                 "uom_id": cls.uom_unit.id,
-                "default_code": "C",
+                "default_code": "C2",
                 "route_ids": [(6, 0, buy_route.ids)],
             }
         )


ddmrp/tests/common.py
```diff
@@ -217,7 +217,7 @@ def setUpClass(cls):
                 "name": "Product C",
                 "is_storable": True,
                 "uom_id": cls.uom_unit.id,
-                "default_code": "C",
+                "default_code": "C2",
```
sebalix, line 220: Not sure this is the right one to fix. It should be fixed in the other private module no?

henrybackman: Changing the code in the private module instead did not fix the issue with the tests but I'm not 100% sure why it didn't

sebalix:
Closing, the issue was similar to the previous fix #532 but coming from another module.

PR #534, opened 2025-05-12
[17.0][IMP] ddmrp: consider that done quantity and demand can be different in stock.move

DavidJForgeFlow:
Since v17, `quantity` field represents the actual done quantity and can differ from `product_qty` which remeains as the original demand when validating a transfer without a backorder. This need to be considered by ddmrp.

Backport from #528 

Depends on https://github.com/OCA/stock-logistics-workflow/pull/1939


LoisRForgeFlow (approved):
:+1:

PR #539, opened 2025-06-11
[17.0][REF] ddmrp_exclude_moves_adu_calc_sales: make boolean compute for robustness

LoisRForgeFlow:
The previous approach was error prone as there could be corner cases in which the `exclude_from_adu` field in SO and moves is misaligned. With a compute and inverse methods approach we ensure that there are no misalignments and keep the field in the stock.move as the only source of truth.


DavidJForgeFlow (approved):
Makes sense and LGTM!

PR #550, opened 2025-07-17
[18.0] ddmrp: fix references to tree views

simahawk:
Fix Uncaught Promise > View types not defined tree found in act_window action undefined.


LoisRForgeFlow (approved):
:+1:

PR #560, opened 2025-09-24
[18.0][PERF] ddmrp_history: Improve domain of multicompany ir rule

grindtildeath:
Field company_id on ddmrp.history is defined as a related to buffer_id.company_id which is a required field on stock.buffer.

Therefore, we do not need to consider the possibility of a null value in the rule, which will in turn improve how Odoo's ORM will generate the SQL to fetch records linked to a single buffer.

In a table with >30M records and 2000 linked to a single buffer, execution time went from 10-15ms to 3-5ms to get all the records from a single buffer.

### DIFF
--- a/ddmrp_history/security/security.xml
+++ b/ddmrp_history/security/security.xml
@@ -6,8 +6,6 @@
         <field name="name">DDMRP History Multi-Company</field>
         <field name="model_id" ref="model_ddmrp_history" />
         <field eval="True" name="global" />
-        <field
-            name="domain_force"
-        >['|',('company_id', 'in', company_ids),('company_id','=',False)]</field>
+        <field name="domain_force">[('company_id', 'in', company_ids)]</field>
     </record>
 </odoo>


twalter-c2c (approved):
Good to know, thanks!

simahawk (approved):

simahawk:
@LoisRForgeFlow ping :)

sebalix (approved):

ddmrp_history/security/security.xml
```diff
@@ -6,8 +6,6 @@
         <field name="name">DDMRP History Multi-Company</field>
         <field name="model_id" ref="model_ddmrp_history" />
         <field eval="True" name="global" />
-        <field
-            name="domain_force"
-        >['|',('company_id', 'in', company_ids),('company_id','=',False)]</field>
+        <field name="domain_force">[('company_id', 'in', company_ids)]</field>
```
LoisRForgeFlow, line 9: @grindtildeath It makes sense and LGTM. Just out of curiosity. Would `[('company_id', 'in', company_ids + [False])]` be more performant than `['|',('company_id', 'in', company_ids),('company_id','=',False)]`? Could you check that?

grindtildeath: @LoisRForgeFlow It was actually the first thing I tried (since I did also see this syntax used in Odoo) but the performance was mostly the same or even slightly worse than with the OR operator in domain.

Unfortunately I didn't save the execution plan after settling on the actual solution after realizing company_id was mandatory anyway.

LoisRForgeFlow: Ok, thanks for the insights!

PR #564, opened 2025-10-10
[18.0][FIX] ddmrp: buffer procure date planned

### LINKED ISSUE #563: Buffer Procurement Wizard planned date wrong for manufactured buffer
In the case of a manufactured buffer, the planned date computation depends on the bom context location. Changing the context location on the bom shows how the dlt is affected.

When we procure manually a buffer with the wizard, it prepares the lines based on the location of the buffer. However, the planned date is based on the "Stock" location dlt instead of that procurement location.

It looks to me that when `buffer._get_date_planned()` is called in ddmrp/wizards/make_procurement_buffer.py
https://github.com/OCA/ddmrp/blob/8a98e41bfb94f5dc1733fb8061de5de610025545/ddmrp/wizards/make_procurement_buffer.py#L33 
it's missing to put the location in the context
https://github.com/OCA/ddmrp/blob/8a98e41bfb94f5dc1733fb8061de5de610025545/ddmrp/models/mrp_bom.py#L66

The consequence, in my case, is that the planned date is using the supplier lead time instead of the dlt of buffered components. On Stock it's purchased, on the production location it is buffered.

cc @LoisRForgeFlow @BernatPForgeFlow @JordiBForgeFlow @meredith-alana @Highcooley

jbaudoux:
In case of manufactured buffer, use the buffer location to compute the planned date.

Closes https://github.com/OCA/ddmrp/issues/563

cc @LoisRForgeFlow 

Tested manually in the interface

### DIFF
--- a/ddmrp/models/mrp_bom.py
+++ b/ddmrp/models/mrp_bom.py
@@ -1,4 +1,5 @@
 # Copyright 2017-24 ForgeFlow S.L. (http://www.forgeflow.com)
+# Copyright 2025 Jacques-Etienne Baudoux (BCIM) <je@bcim.be>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
 import logging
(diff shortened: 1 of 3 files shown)


LoisRForgeFlow (commented):
I'm unable to reproduce the issue, my answer is directly in the issue to keep the problem discussion there. https://github.com/OCA/ddmrp/issues/563#issuecomment-3401099215

PR #565, opened 2025-10-10
[18.0][FIX] ddmrp: bom line buffered

jbaudoux:
Consider a bom line as buffered if there is a buffer on a child location

Use case:
* Have pick that sources products from several locations under "Production" location
* Have a manufactured buffer on that "Production" location
* Have distributed buffers for the components on the "Production/Shelf" location

The bom should consider the bom lines as buffered. This has an impact on the dlt computation.

Components are located on different areas under "Production" and the buffer is on the area where the component must be replenished

cc @LoisRForgeFlow @BernatPForgeFlow 

Manually tested


LoisRForgeFlow (approved):
LGTM

PR #568, opened 2026-01-06
[18.0][FIX] ddmrp: Use supplier from product template for vendor code compute

grindtildeath:
In case the product set on the buffer only has a supplier info set at product template level and not on the product variant itself, the product's vendor code was not set on the buffer.


LoisRForgeFlow (approved):
LGTM :+1:

PR #571, opened 2026-01-23
[16.0][FIX] ddmrp_sale: can_server_sale in multi step deliveries + cancelled moves

DavidJForgeFlow:
Adapt the computation of can_serve_sale as there are some cases in which the buffer cannot be detected if, for example, the sale route is in more than 1 step. We do not use the method in stock_helper as it disappears in future versions and we don't want to add dependency to the new module.

When a quotation is canceled and  its stock.moves are canceled, so we can take into account quotations with moves (as they must be canceled). You cannot cancel a sale order if it has done moves.

### DIFF
--- a/ddmrp_sale/tests/test_ddmrp_sale.py
+++ b/ddmrp_sale/tests/test_ddmrp_sale.py
@@ -39,6 +39,9 @@ class TestDDMRPSale(TestDdmrpCommon):
     def test_01_can_serve_sales(self):
         self.assertTrue(self.buffer_a.can_serve_sales)
         self.assertFalse(self.buffer_internal.can_serve_sales)
+        self.warehouse.delivery_steps = "pick_pack_ship"
+        self.buffer_a._compute_can_serve_sales()
+        self.assertTrue(self.buffer_a.can_serve_sales)
 
     def test_02_sales_quotation_included_as_demand(self):
         self._refresh_involved_buffers()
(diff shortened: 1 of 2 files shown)


ddmrp_sale/tests/test_ddmrp_sale.py
```diff
@@ -116,3 +116,60 @@ def test_03_sol_uom(self):
...
+        )
+        # Buffer A sees quotations because it can serve SO's.
+        diff = self.buffer_a.qualified_demand - self.buffer_internal.qualified_demand
+        self.assertEqual(diff, 17)
+        so.action_confirm()
+        self.assertTrue(so.order_line.move_ids)
+        self._refresh_involved_buffers()
+        self.assertEqual(
+            self.buffer_a.qualified_demand, self.buffer_internal.qualified_demand
+        )
```
LoisRForgeFlow, line 166: @DavidJForgeFlow Could you also check that the increase in the internal buffer NFP is 17 comparing before and after validation? It is now not clear if the buffer A NFP decreased or the internal one increased.

DavidJForgeFlow: Added the check that ensures the nfp is equal before and after validation in Buffer A and in Buffer Internal the NFP increases in 17

LoisRForgeFlow (approved):
LGTM

PR #572, opened 2026-01-26
[16.0][IMP] ddmrp: add search hook

DavidJForgeFlow:
When computing the ADU, a hook is added to get the domain of buffers to update.


LoisRForgeFlow (changes requested):
Backport https://github.com/OCA/ddmrp/pull/562 instead

PR #148, opened 2021-06-23
[13.0][IMP] ddmrp add product vendor code

TDu:
Add the product vendor code on the stock buffer based on the already
computed main supplier.

PR #575, opened 2026-02-05
[16.0][IMP] ddmrp_product_replace

DavidJForgeFlow:
It was reported that some users manually modified the replaced by manually. This was not intended so we force the user to use the wizard.

### DIFF
--- a/ddmrp_product_replace/views/stock_buffer_view.xml
+++ b/ddmrp_product_replace/views/stock_buffer_view.xml
@@ -39,8 +39,19 @@
                     />
                 </button>
             </div>
-            <field name="product_vendor_code" position="after">
-                <field name="replaced_by_id" />
+            <field name="product_id" position="after">
+                <label for="replaced_by_id" />
+                <div name="replaced_by_id" class="o_row">
+                    <field name="replaced_by_id" class="oe_inline" />
+                    <button
+                        title="Replace Buffer"
+                        name="%(action_ddmrp_product_replace_wizard)d"
+                        icon="fa-edit"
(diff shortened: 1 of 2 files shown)


ddmrp_product_replace/views/stock_buffer_view.xml
```diff
@@ -40,7 +40,18 @@
                 </button>
             </div>
             <field name="product_vendor_code" position="after">
-                <field name="replaced_by_id" />
+                <label for="replaced_by_id" />
```
LoisRForgeFlow: @DavidJForgeFlow Actually, can you also relocate the field below the product_id field in the form view? It was there before but in [this](https://github.com/OCA/ddmrp/pull/148/changes#diff-ddfa91fda93d177593dbd854e80ca3179f7744c1e09c14f3d72d4a0e6ee9d714L38) unrelated change it was moved. It is maybe not ideal in the header, but in the Procurement setting section it doesn't make any sense.

DavidJForgeFlow: Changed!

LoisRForgeFlow (approved):
LGTM 👍

ddmrp_product_replace/views/stock_buffer_view.xml
LoisRForgeFlow: thanks!

LoisRForgeFlow:
@DavidJForgeFlow could you fwport this?

DavidJForgeFlow:
Forward ports done!

PR #577, opened 2026-02-12
[18.0][IMP] ddmrp_product_replace

DavidJForgeFlow:
It was reported that some users manually modified the replaced by manually. This was not intended so we force the user to use the wizard.

Forward port of #575


LoisRForgeFlow (approved):
Tested on runboat 👍

PR #591, opened 2026-03-05
[18.0][FIX] ddmrp: compute outgoing_reservation_qty using the quantity in the product unit of measure

JordiBForgeFlow:
@LoisRForgeFlow


LoisRForgeFlow:
Added a test case.

LoisRForgeFlow (approved):
/ocabot merge patch

LoisRForgeFlow:
Backported to 17.0, do not apply it in 16.0 because 16.0 uses reserved_qty. Basically this was a migration mistake in 17.0

PR #594, opened 2026-03-23
[16.0][FIX] ddmrp: hide misleading ADU details buttons when not relevant

LoisRForgeFlow:
Show only indirect demand button when the method is going to use it.

Example:
image

image


PaulaMForgeFlow (approved):
LGTM

PR #608, opened 2026-04-30
[18.0][FIX] ddmrp_report_part_flow_index: do not include archived buffers in the report

LoisRForgeFlow:
Also, take the ocassion to make buffer_id clickable in tree view.


PaulaMForgeFlow (approved):
Functional review ok

PR #609, opened 2026-05-13
[19.0][FIX] ddmrp: skip execution chart compute when no product is set

PaulaMForgeFlow:
Fix to solve error when creating buffer.


LoisRForgeFlow (approved):
Thanks for the patch! 👍

PR #612, opened 2026-06-12
[19.0][FIX] ddmrp: aggregate future actual demand in product UoM

dnplkndll:
`_calc_adu_future_demand` sums `product_uom_qty:sum` — the move-UoM "Demand" field — so a future delivery of 5 Dozen contributes 5 instead of 60 to the ADU, undersizing buffers for any secondary-UoM demand.

18.0 summed `product_qty` (product UoM, still stored/groupable in 19.0), as this method has since the 13.0 refactor; every other quantity ddmrp aggregates is product-UoM-normalized (past ADU via `quantity_product_uom`, qualified demand, incoming).

The regression test alone against unmodified 19.0 fails ([run](https://github.com/ledoent/ddmrp/actions/runs/27344118311/job/80788172608), 1 failed of 61):

```
FAIL: TestDdmrp.test_51_adu_calculation_future_actual_move_uom
AssertionError: 0.042 != 0.5
```

With the one-line fix it passes ([run](https://github.com/ledoent/ddmrp/actions/runs/27344116959/job/80788167832)).

### DIFF
--- a/ddmrp/models/stock_buffer.py
+++ b/ddmrp/models/stock_buffer.py
@@ -1550,7 +1550,7 @@ class StockBuffer(models.Model):
         elif self.adu_calculation_method.source_future == "actual":
             domain = self._future_moves_domain(date_from, date_to, locations)
             for [total] in self.env["stock.move"]._read_group(
-                domain, aggregates=["product_uom_qty:sum"]
+                domain, aggregates=["product_qty:sum"]
             ):
                 qty += total or 0.0
         return qty / horizon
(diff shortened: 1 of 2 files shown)


LoisRForgeFlow (approved):
Thanks for the fix, it is correct indeed. I drop here a picture with a simple schema to make it clear for anyone with doubts:



In this case product_qty is correct because in the future domain we include non-done moves so we need to use the "Demand", in the past domain quantity_product_uom is needed because that was the actually moved quantity.

LoisRForgeFlow:
Thanks for the test and the red-proof BTW, that saves me time reviewing :+1: 

/ocabot merge patch

LoisRForgeFlow:
No need to backport this patch, it was a bug introduced during migration to 19.0, it is correct in 18.0: https://github.com/OCA/ddmrp/blob/5e4cd374153339fbf63572debee80437df595ec3/ddmrp/models/stock_buffer.py#L1545

PR #613, opened 2026-06-12
[19.0][FIX] ddmrp: NewId-safe buffer-link lookup on purchase lines

dnplkndll:
One-character fix (`rec.id` → `rec.ids`) in `_find_buffer_link`'s move search, plus a regression test.

The method is reachable on unpersisted (NewId) lines — via the `_product_id_change` onchange helper and 19.0's form-view validation — and searching with `rec.id` logs `Domains don't support NewId`, which fails OCA checklog for any dependent module that loads a purchase view or creates an order in tests. `rec.ids` is empty for NewId records, keeping the lookup a no-op there (same effective behavior as before).

The test drives a purchase line through `Form` under `assertNoLogs("odoo.domains")` — it fails on the unmodified branch.

### DIFF
--- a/ddmrp/models/purchase_order.py
+++ b/ddmrp/models/purchase_order.py
@@ -81,7 +81,7 @@ class PurchaseOrderLine(models.Model):
         move_model = self.env["stock.move"]
         for rec in self.filtered(lambda r: not r.buffer_ids):
             mto_move = move_model.search(
-                [("created_purchase_line_ids", "in", rec.id)], limit=1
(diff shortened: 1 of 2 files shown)


LoisRForgeFlow (commented):
@dnplkndll After merging your first patch, we have a merge conflict in this PR, can you solve it? Thanks!

LoisRForgeFlow (approved):
Thanks 👍

LoisRForgeFlow:
Warning does not appear in v18 apparently, so no need to backport.

PR #614, opened 2026-06-12
[19.0][FIX] ddmrp: pin test_04 ADU-window test to a fixed weekday

dnplkndll:
Makes `test_04_adu_calculation_window_past_calendar` deterministic regardless of which weekday CI runs on.

The test anchors its demand moves on `datetime.today()` through `resource.calendar.plan_days()` (working-day-aware). On weekend runs the working-day window shifts by a day, so one in-window move drops out and the ADU assertion fails (`6.667 != 8.333`). `_calc_adu` also reads the real current date, so freezing the test data alone is not enough — the whole method is pinned with `freeze_time` (already a `ddmrp` test dependency, cf. `test_distributed_max_proc_time.py`).

Not visible on the current `19.0` branch because its last CI run landed on a weekday; it surfaces on Saturday/Sunday runs.

Evidence from a Sunday CI run: [failing on the unmodified branch](https://github.com/ledoent/ddmrp/actions/runs/27092429455/job/79958330677) (`FAIL: test_04 ... AssertionError: 6.667 != 8.333`) vs [green with the fix, same day](https://github.com/ledoent/ddmrp/actions/runs/27095225243/job/79966030582).

### DIFF
--- a/ddmrp/tests/test_ddmrp.py
+++ b/ddmrp/tests/test_ddmrp.py
@@ -5,6 +5,8 @@
 import json
 from datetime import datetime, time, timedelta
 
+from freezegun import freeze_time
+
 from odoo import fields
 from odoo.exceptions import ValidationError
 from odoo.tests import Form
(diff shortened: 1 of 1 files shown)


LoisRForgeFlow (approved):
Definitely, thanks! 👍

PR #622, opened 2026-06-18
[19.0][FIX] ddmrp: guard stock.buffer references in shared views

DavidJForgeFlow:
Fields and buttons referencing stock.buffer were injected into views of models that can be opened without stock.group_stock_user (product, purchase). Reading the related buffer records raised an AccessError for users without that group. Restrict those elements with groups="stock.group_stock_user".

### DIFF
--- a/ddmrp/views/product_view.xml
+++ b/ddmrp/views/product_view.xml
@@ -15,6 +15,7 @@
                     class="oe_stat_button"
                     icon="fa-flask"
                     invisible="buffer_count == 0"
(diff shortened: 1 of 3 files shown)


LoisRForgeFlow (approved):

LoisRForgeFlow:
@DavidJForgeFlow does this need backport?

DavidJForgeFlow:
Yes, this needs to be backported. Can anyone do it?

LoisRForgeFlow:
It would be great if you can do it.

PR #629, opened 2026-07-03
[16.0][FIX] ddmrp: guard stock.buffer references in shared views

DavidJForgeFlow:
Fields and buttons referencing stock.buffer were injected into views of models that can be opened without stock.group_stock_user (product, purchase). Reading the related buffer records raised an AccessError for users without that group. Restrict those elements with groups="stock.group_stock_user".

BP of https://github.com/OCA/ddmrp/pull/622


LoisRForgeFlow (approved):
/ocabot merge minor

PR #633, opened 2026-07-03
[15.0][UPD] copier: 1.43

DavidJForgeFlow:


LoisRForgeFlow (approved):
Tested locally 👍  15.0 is no longer maintained but anyway, let's update in case we need to merge some important bugfix.

PR #634, opened 2026-07-10
[19.0][FIX] ddmrp: exclude make-to-order demand from the net flow position

JordiBForgeFlow:
Qualified demand counted every outgoing stock move for the product, including make-to-order (pegged) demand. MTO demand is served by its own dedicated supply, not from the buffer, so counting it deflated the net flow position and led to over-procurement of MTS buffers.

This was also inconsistent with the supply side, which already keeps MTO purchase lines out of MTS buffers (purchase.order.line._find_buffer_link).

Use one shared criterion on both sides, based on the same relation (stock.move.created_purchase_line_ids <-> purchase.order.line.move_dest_ids):

- stock.move._ddmrp_is_mto(): the move raised a purchase to satisfy itself.
- purchase.order.line._ddmrp_is_mto(): the line was raised for a demand move.

_search_stock_moves_qualified_demand now drops MTO moves, and _find_buffer_link is refactored to use the mirrored helper. Because both ends test the same relation, a pegged demand move and the purchase line it created are always excluded together, keeping the net flow position consistent.

### DIFF
--- a/ddmrp/__manifest__.py
+++ b/ddmrp/__manifest__.py
@@ -5,7 +5,7 @@
 {
     "name": "DDMRP",
     "summary": "Demand Driven Material Requirements Planning",
-    "version": "19.0.1.1.0",
+    "version": "19.0.1.1.1",
     "license": "LGPL-3",
     "development_status": "Production/Stable",
     "author": "ForgeFlow, Odoo Community Association (OCA)",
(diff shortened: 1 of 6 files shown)


LoisRForgeFlow (approved):
It has a manual bump, but otherwise the fix looks fine.

PR #644, opened 2026-09-05
[FIX] ddmrp: prevent duplicate replenishment on half-refreshed net flow position

divad1196:
## Short Explanation
We have 2 kind of jobs: in and out that recomputes only the incoming or only the outgoing quantities.
But each of these jobs can trigger a replenishment.

→ if the first job triggers the replenishment, the buffer’s nfp values are partially outdated

## Fix

We must always compute both `in` and `out` values (at least when we call `do_auto_procure` method)

NOTE:
This should not impact the performance too much, we usually had both in and out jobs in the queue anyway.
Now we will only have 1 job doing both. The deduplication prevent running it twice.

### DIFF
--- a/ddmrp/models/stock_buffer.py
+++ b/ddmrp/models/stock_buffer.py
@@ -2102,12 +2102,14 @@ class StockBuffer(models.Model):
         if not only_nfp or only_nfp == "in":
             self._calc_incoming_dlt_qty()
         self._calc_net_flow_position()
-        self._calc_distributed_source_location()
         self._calc_planning_priority()
         self._calc_execution_priority()
         self.mrp_production_ids._calc_execution_priority()
         self.mapped("purchase_line_ids")._calc_execution_priority()
         if not only_nfp:
+            # The source location only changes when the stock rules change, so
+            # walking them on every partial refresh is a pure cost.
+            self._calc_distributed_source_location()
             # re-compoute red to force in cascade the recalculation of zones.
             self._compute_red_zone()
         self.do_auto_procure()
--- a/ddmrp/models/stock_move.py
+++ b/ddmrp/models/stock_move.py
@@ -81,34 +81,23 @@ class StockMove(models.Model):
         # outgoing and `in_buffers` as incoming.
         out_buffers = in_buffers = self.env["stock.buffer"]
         for move in self:
-            out_buffers |= move.mapped("product_id.buffer_ids").filtered(
-                lambda buffer, move=move: (
-                    move.location_id.is_sublocation_of(buffer.location_id)
-                    and (
-                        not move.location_dest_id.is_sublocation_of(buffer.location_id)
-                        or (
-                            move.location_final_id
-                            and not move.location_final_id.is_sublocation_of(
-                                buffer.location_id
-                            )
-                        )
-                    )
+            for buffer in move.product_id.buffer_ids:
+                location = buffer.location_id
+                src_in_buffer = move.location_id.is_sublocation_of(location)
+                dest_in_buffer = move.location_dest_id.is_sublocation_of(location)
+                has_final = bool(move.location_final_id)
+                final_in_buffer = (
+                    has_final and move.location_final_id.is_sublocation_of(location)
                 )
-            )
-            in_buffers |= move.mapped("product_id.buffer_ids").filtered(
-                lambda buffer, move=move: (
-                    not move.location_id.is_sublocation_of(buffer.location_id)
-                    and (
-                        move.location_dest_id.is_sublocation_of(buffer.location_id)
-                        or (
-                            move.location_final_id
-                            and move.location_final_id.is_sublocation_of(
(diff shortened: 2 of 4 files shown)


divad1196:
@jbaudoux

jbaudoux:
From what I remember, nfp_out is when situation changed at buffer location, so it should compute:
* self._calc_qualified_demand()
* self._calc_incoming_dlt_qty()

nfp_in is when situation changed at source location, so it should compute:
* self._calc_distributed_source_location()

and that is usefull to compute only if you limit to available stock at source location

divad1196:
@jbaudoux from the code analysis, both relate to the buffer location:
- `only_nfp=out`: recompute only what is going **out** of the buffer location
- `only_nfp=in`: recompute only what is going **inside** of the buffer location

[_update_ddmrp_nfp](https://github.com/OCA/ddmrp/blob/18.0/ddmrp/models/stock_move.py#L114) pass the `only_nfp="in"` flag to the moves whose's location_dest_id is the buffer location.

The method [_find_buffers_to_update_nfp](https://github.com/OCA/ddmrp/blob/18.0/ddmrp/models/stock_move.py#L98) that returns the `out_buffers` and `in_buffers` only check the buffer's `location_id`, never it's source.

ddmrp/models/stock_buffer.py
```diff
@@ -2102,15 +2100,23 @@ def cron_actions(self, only_nfp=False):
...
             # re-compoute red to force in cascade the recalculation of zones.
             self._compute_red_zone()
-        self.do_auto_procure()
+        return True
+
+    def cron_actions(self, only_nfp=False):
+        """This method is meant to be inherited by other modules in order to
+        enhance extensibility."""
+        self._recompute_nfp(only_nfp=only_nfp)
+        self.do_auto_procure(skip_nfp_refresh=not only_nfp)
```
divad1196, line 2119: Need to change this.

Either we remove `only_nfp` from the function, or we must skip `do_auto_procure` completely

ddmrp/models/stock_move.py
```diff
@@ -76,49 +62,40 @@ def create(self, vals_list):
...
-            buffer.cron_actions(only_nfp="out")
-        for buffer in in_buffers.with_context(no_ddmrp_history=True):
-            buffer.cron_actions(only_nfp="in")
+        out_buffers, in_buffers = self._find_buffers_affected_by_moves()
+        # Procuring is irreversible, so it must never weigh one fresh figure
+        # against one inherited: refresh both sides before deciding.
+        for buffer in (out_buffers | in_buffers).with_context(
+            no_ddmrp_history=True
+        ):
+            buffer.cron_actions()
```
divad1196, line 98: NOTE: with `ddmrp_cron_actions_as_job`, the calls to `cron_actions` become jobs

This is the core of the fix: we always recompute `in` and `out` for the buffers.

ddmrp/models/stock_move.py
```diff
@@ -76,49 +62,40 @@ def create(self, vals_list):
...
 
     def _update_ddmrp_nfp(self):
         if self.env.context.get("no_ddmrp_auto_update_nfp"):
             return True
-        out_buffers, in_buffers = self._find_buffers_to_update_nfp()
-        for buffer in out_buffers.with_context(no_ddmrp_history=True):
-            buffer.cron_actions(only_nfp="out")
-        for buffer in in_buffers.with_context(no_ddmrp_history=True):
-            buffer.cron_actions(only_nfp="in")
+        out_buffers, in_buffers = self._find_buffers_affected_by_moves()
```
divad1196, line 92: NOTE:
renamed the method because it was confusing IMO

`_find_buffers_to_update_nfp` sounds like it will search any buffer that needs to be updated, but it only finds the one related to the moves.

jbaudoux (approved):

ddmrp/models/stock_buffer.py
```diff
@@ -2084,9 +2088,8 @@ def refresh_buffer(self):
         self.cron_actions(only_nfp=False)
         return True
 
-    def cron_actions(self, only_nfp=False):
-        """This method is meant to be inherited by other modules in order to
-        enhance extensibility."""
+    def _recompute_nfp(self, only_nfp=False):
```
divad1196, line 2091: NOTE: This function is called `_recompute_nfp` to be consistent with `_update_ddmrp_nfp`

`_update_ddmrp_nfp` -> `cron_actions` -> `_recompute_nfp`.

LoisRForgeFlow (changes requested):
Hi @divad1196 

Can you keep this PR to the minimum needed for the fix? I mean the tests and the actual fix. I prefer if you propose the method renames and rearrangement in a different PR because that has implication for existing installations, whereas the simple patch can be merged more easily.

Thanks!

divad1196:
@LoisRForgeFlow I have removed the cosmetic changes

